PR #1, opened 2017-04-07
get preview.pdf working with http auth

leohemsted:
This required a few things
* loading VCAP_SERVICES, which contains the token required to access
  this endpoint
* updating the makefile to set VCAP_SERVICES when running locally
* setting up basic flask app structure with blueprints
* installing flask-httpauth to manage this


leohemsted:
superseded by #2

PR #2, opened 2017-04-11
Get template previews working

leohemsted:
🎉 

Please run locally using nothing but makefile commands! Let me know if anything breaks cos of my local env


quis:
Couple of things I’ve found:
- Need to add `admin_base_url` to the scheme (getting error `Additional properties are not allowed ('admin_base_url' was unexpected)`)
- Schema requires template to have `id` and `name`, do we actually use these?
- Had to add `CF_SPACE ?= sandbox` to line 21 of `Makefile` to get it working locally

There’s also the issue of fonts. Probably not for this PR, but need to install them in the container or look at the possibility of using webfonts. Here is [what Andras had to do to install the fonts on the admin boxes](https://github.com/alphagov/notifications-aws/commit/ea8cc27d4a8159f0efe503a93b0b2bf63cfff658).

quis:
Tests passing when running in Docker on my machine.

requirements.txt
```diff
@@ -6,10 +6,12 @@
 # app requirements
 Flask==0.12.1
 Flask-WeasyPrint==0.5
+Flask-HTTPAuth==3.2.2
 html5lib==1.0b10
 wand==0.4.4
+jsonschema==2.6.0
 
-git+https://github.com/alphagov/notifications-utils.git@13.10.0#egg=notifications-utils==13.10.0
+git+https://github.com/alphagov/notifications-utils.git@15.0.2#egg=notifications-utils==15.0.2
```
quis: 15.1.2 is the latest version now.

quis:
Only other thing is the PNG conversion looks a bit muddier, but can probs dig into that separately.

quis:
PNG | PDF
---|---
image | image

README.md
```diff
@@ -58,16 +58,20 @@ curl \
...
+      "subject": "foo", \
+      "content": "bar" \
+    }, \
     "values":null, \
-    "letter_contact_block":null, \
-    "admin_base_url":"http://localhost:6012" \
+    "letter_contact_block":null \
   }' \
-  localhost:6013/preview.pdf
+  https://localhost:6013/preview.pdf
```
quis: Should be `http`

PR #6, opened 2017-04-18
Clarify setup instructions

kentsanggds:
Whilst getting it to work I hit some issues, so I've updated the README to hopefully make it clearer for the next person

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ make (sandbox|preview|staging|production) cf-deploy
 
 During development it may be preferable to run locally.
 
-If you haven't created the virtualenv follow these steps - 
+If you haven't installed the app yet follow these steps - 
 
 ```shell
 # binary dependencies
@@ -82,9 +82,10 @@ brew install imagemagick ghostscript cairo pango
 
 mkvirtualenv -p /usr/local/bin/python3 notifications-template-preview
 pip install -r requirements.txt
-```
 
-Then create a `version.py` file under the app folder, you can copy `version.py.dist` to `version.py` to get it running.
+# create a version file
+make _generate-version-file
+```
 
 You will also need to set VCAP_SERVICES (extracted from Makefile) -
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,4 +84,22 @@ mkvirtualenv -p /usr/local/bin/python3 notifications-template-preview
...
+
+```shell
+./scripts/run_app.sh 6013
+```
+
+Thereafter activate the virtualenv prior to executing the run script -
+
+```shell
+workon notifications-template-preview
+```
```
leohemsted: dont see why this is separate to the block above - since it needs to be run first

leohemsted: in fact i think all three of these code blocks should be combined since it doesn't make sense to do any one step without the others

README.md
```diff
@@ -89,4 +84,22 @@ mkvirtualenv -p /usr/local/bin/python3 notifications-template-preview
 pip install -r requirements.txt
 ```
 
-Then to run, activate the virtualenv and call `./scripts/run_app.sh 6013`
+Then create a `version.py` file under the app folder, you can copy `version.py.dist` to `version.py` to get it running.
```
leohemsted: does `make _generate-version-file` work?

README.md
```diff
@@ -79,7 +72,9 @@ make (sandbox|preview|staging|production) cf-deploy
 
 ## Running locally
 
-During development it may be preferable to run locally - in which case you'll need to run the following steps
+During development it may be preferable to run locally.
+
+If you haven't created the virtualenv follow these steps -
```
leohemsted: hmm if we're cleaning up the file, then this should probably have two sub-sub-sub headings - `first time installation` and `running the app` or similar, so it's apparent that you only need to install once. then we don't need to say "if you haven't created the virtualenv"

README.md
kentsanggds: Is it meant to run it automatically? I just manually created the version.py file after looking at the errors

leohemsted: `make _generate-version-file` will run automatically... if you run one of the commands that depends on it. this is designed to be run using the makefile commands to control it through docker, it's run in docker in the real world, using a different version of python (3.6) to everything else, using particular versions of binary dependencies that aren't going to be the same as what you get on brew - i wouldn't really recommended running locally, i just threw in those lines for the curious

README.md
kentsanggds: Running mkvirtualenv actually puts you in the virtualenv so I've added this as a separate block

leohemsted: yes, but you'll only run `mkvirtualenv` once - the same time as you install binary dependencies, install python, etc. you'll run `workon` every time you want to edit the template-preview app

README.md
kentsanggds: ok, but it's still useful to just let people know what needs to be done to run it locally, could change the documentation to use make _generate-version-file instead

README.md
kentsanggds: yes, I think that's clear in the notes that `mkvirtualenv` is only called if a virtualenv hasn't been created, thus prior to this code block there's this comment -

_Thereafter activate the virtualenv prior to executing the run script -_

I read it through as initial local setup and then re-running the app locally

PR #7, opened 2017-04-19
Allow PNGs to be generated for any page of a letter

quis:
Adds two things:
- an endpoint for the frontend to get the count of pages in a letter
- a query parameter that the frontend can use to specify a page number, when it’s asking for a `preview.png`

Hacking the admin app, it gives us something like this:

***

screen shot 2017-04-19 at 13 09 07

***

The idea being that we can show people every page of a letter in the admin app, without having to open the PDF.

### DIFF
--- a/app/preview.py
+++ b/app/preview.py
@@ -39,15 +39,23 @@ def validate_preview_request(json):
         abort(400, exc)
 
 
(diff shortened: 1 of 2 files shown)


minglis:
Seems OK.

👍

PR #8, opened 2017-04-20
Allow logo to be configurable in template previews

quis:
Takes `dvla_org_id` in the post body (which is what the admin app knows about) and maps it to the filename of the logo (which is encapsulated within this app).

This will let us preview letters with the Land Registry logo.


app/preview.py
```diff
@@ -77,7 +78,8 @@ def view_letter_template(filetype):
         values=json['values'] or None,
         contact_block=json['letter_contact_block'],
         # we get the images of our local server to keep network topography clean, which is just http://localhost:6013
-        admin_base_url='http://localhost:6013'
+        admin_base_url='http://localhost:6013',
+        logo_file_name=current_app.config['LOGO_FILENAMES'][json['dvla_org_id']],
```
leohemsted: reckon we should `abort(400)` if this isn't specified (rather than 500 KeyError)

quis: Good point. Changed, added a test and squashed into here: https://github.com/alphagov/notifications-template-preview/pull/8/commits/f4b52258870b5cba65b86540afb2cf4cbbc75bed#diff-715e0c95f956ae6eead032ba5491f2f2R83

PR #11, opened 2017-04-26
Upgrade memory to fix issue with previewing letters

imdadahad:
This upgrades the memory on the image so that memory doesn't run low and cause the letter previews to fail.


quis:
Going to just merge this so no-one accidentally deploys template preview without it and reintroduces the bug.

PR #12, opened 2017-05-03
Build libcairo by hand

quis:
The version of libcairo that comes with Debian stable is quite old. It gives some warnings about ‘known rendering problems’, which might be why logos in our image previews look a bit crusty.

This commit adds step to the `Dockerfile` to build a newer version of Cairo from source, rather than use Apt.

N.B. I have not yet found a way to check if our Python apps are using this version, or still using the old one. So, while this does run, consider it a WIP. It doesn’t seem to have changed anything yet.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,11 +25,21 @@ RUN \
 	&& apt-get update \
 	&& apt-get install -y --no-install-recommends \
 		libpango1.0-dev \
-		libcairo2-dev \
 		libmagickwand-dev \
 		ghostscript \
 		imagemagick \
 		gsfonts
+
+RUN \
+	echo "Install newer version of Cairo" \
+	&& curl -O https://www.cairographics.org/releases/cairo-1.14.8.tar.xz \
+	&& tar xf cairo-1.14.8.tar.xz \
+	&& cd cairo-1.14.8 \
+	&& ./configure \
+	&& make \
+	&& make install \
+	&& cd .. \
+	&& rm -r cairo-1.14.8 cairo-1.14.8.tar.xz
 
 RUN \
 	echo "Clean up" \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -25,13 +25,23 @@ RUN \
...
 RUN \
+	echo "Install newer version of Cairo" \
+	&& curl -O https://www.cairographics.org/releases/cairo-1.14.8.tar.xz \
+	&& tar xf cairo-1.14.8.tar.xz \
+	&& cd cairo-1.14.8 \
+	&& ./configure \
+	&& make \
+	&& make install \
+	&& cd .. \
+	&& rm -r cairo-1.14.8 cairo-1.14.8.tar.xz
```
leohemsted: tthis could do with a bit of cleanup - suggestions

* can curl without `-O`, then pipe the result straight into `curl {url} | tar x`
* best practice is not to use `cd` in RUN commands, use `WORKDIR` instead - https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#workdir 
  - then `WORKDIR /` after to go back
* i'd possibly `WORKDIR /tmp/` before this, that way we don't need the `rm` step as we remove `/tmp/*` in the next step

leohemsted:
i think to get the version dynamically we'd need to create a small C script that included the dynamic cairo lib then called `cairo_version()` 😓 

https://cairographics.org/manual/cairo-Version-Information.html#cairo-version

leohemsted:
- [ ] https://github.com/alphagov/notifications-template-preview/pull/12

docker/Dockerfile
quis: Cleaned up in c15674e37bc06c93509ea4c1391235f727800c92

quis: Nope, ea63dbf

PR #20, opened 2017-07-04
Use newer version of libcairo without compiling it

quis:
Stumbled across this while trying to get this app running on my new Mac: https://github.com/aquavitae/docker-weasyprint/pull/4

Does what we tried to do in https://github.com/alphagov/notifications-template-preview/pull/12 but actually seems to work (ie running the tests no longer gives warnings about an outdated version of Cairo).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -31,10 +31,10 @@ RUN \
 		gsfonts
 
 RUN echo "Install libcairo2 from stretch" \
-	&& echo 'APT::Default-Release "stretch";' > /etc/apt/apt.conf \
-	&& mv /etc/apt/sources.list /etc/apt/sources.list.d/stable.list \
+	&& echo 'APT::Default-Release "jessie";' > /etc/apt/apt.conf \
+	&& mv /etc/apt/sources.list /etc/apt/sources.list.d/jessie.list \
 	&& echo "deb http://deb.debian.org/debian stretch main" \
-		> /etc/apt/sources.list.d/testing.list
+		> /etc/apt/sources.list.d/stretch.list
 
 RUN apt-get -y update
 RUN apt-get -t stretch install -y libcairo2=1.14.8-1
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -30,19 +30,15 @@ RUN \
...
+	&& echo 'APT::Default-Release "stretch";' > /etc/apt/apt.conf \
+	&& mv /etc/apt/sources.list /etc/apt/sources.list.d/stable.list \
+	&& echo "deb http://deb.debian.org/debian stretch main" \
+		> /etc/apt/sources.list.d/testing.list
+
+RUN apt-get -y update
+RUN apt-get -t stretch install -y libcairo2=1.14.8-1
+RUN apt-get -y clean
 
 WORKDIR /
```
leohemsted: dont need this WORKDIR any more, it was just to get back out of /tmp/

docker/Dockerfile
```diff
@@ -30,19 +30,15 @@ RUN \
...
-
-WORKDIR /tmp/cairo-1.14.8/
-
-RUN \
-  echo "Installing newer version of Cairo" \
-	&& ./configure \
-	&& make \
-	&& make install
+RUN echo "Install libcairo2 from stretch" \
+	&& echo 'APT::Default-Release "stretch";' > /etc/apt/apt.conf \
```
leohemsted: https://github.com/aquavitae/docker-weasyprint/pull/5/files

by the looks of it, this can be `'APT::Default-Release "jessie";'`

then the line below goes from `stable.list` to `jessie.list` and the line below that goes from `testing.list` to `stretch.list`

😵

docker/Dockerfile
quis: Squashed out.

docker/Dockerfile
quis: Changed in b73056c

PR #29, opened 2017-07-20
Image mem problem

venusbb:
Memory from this statement is not freed after the function call
pdf = Image(blob=pdf_endpoint.get_data(), resolution=150)
Reason unknown. Also forum suggested that it may be related to Image.sequence. 
Solution is to put the the pdf in a with statement so that memory is garbaged in __exit__

https://stackoverflow.com/questions/44209861/how-to-reduce-wand-memory-usage

### DIFF
--- a/app/preview.py
+++ b/app/preview.py
@@ -44,25 +44,22 @@ def png_from_pdf(pdf_endpoint, page_number):
     output = BytesIO()
 
     with Image(blob=pdf_endpoint.get_data(), resolution=150) as pdf:
-        image = Image(width=pdf.width, height=pdf.height)
+        with Image(width=pdf.width, height=pdf.height) as image:
+            try:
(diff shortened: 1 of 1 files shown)


app/preview.py
```diff
@@ -42,24 +42,24 @@ def validate_preview_request(json):
...
 
-    image.composite(page, top=0, left=0)
-    converted = image.convert('png')
-    converted.save(file=output)
+            image.composite(page, top=0, left=0)
+            converted = image.convert('png')
+            converted.save(file=output)
 
-    output.seek(0)
+            output.seek(0)
```
quis: This line (and the ones underneath it) don’t need to be inside the `with` blocks.

venusbb:
True. Now moved them out.

PR #54, opened 2017-12-07
Use JPG version of MMO logo

quis:
There is a problem that the logo doesn’t show up when:
- it’s any version of the the MMO logo saved as a PNG
- _and_ there is some text in the letter contact block area of the template (which is why it was appearing to work locally with cURL before, because the cURL command wasn’t including any text for the letter contact block)

We don’t know why this bug is happening, but using the JPG version of the logo that they sent us originally makes it go away.


quis:
Here’s a sample cURL command that replicates the bug with the existing logo:

```shell
curl \
  -O \
  -X POST \
  -H "Authorization: Token my-secret-key" \
  -H "Content-type: application/json" \
  -d '{
    "dvla_org_id": "008",
    "letter_contact_block": "Government Equalities Office\nSanctuary Buildings,\nLondon, SW1P 3BT\n\nwww.geo.gov.uk\n\n((contact number))\n\nM M M M M M M M M M M M M M M M M M M ",    
    "values": null,
    "template": {
      "subject": "Subject H1 Helvetica Neue bold 20/26pt",
      "archived": false,
      "name": "Chris Hill-Scott",
      "updated_at": "2017-03-13T11:17:05.062155+00:00",
      "created_at": "2016-12-05T12:11:29.113331+00:00",
      "redact_personalisation": false,
      "version": 9,
      "reply_to": null,
      "template_type": "letter",
      "template_redacted": "e59ae9e4-befd-44e1-959b-ef4e97aa06ce",
      "service": "42a9d4f2-1444-4e22-9133-52d9e406213f",
      "created_by": "2c45486e-177e-40b8-997d-5f4f81a461ca",
      "process_type": "normal",
      "content": "",            
      "id": "2c45486e-177e-40b8-997d-5f4f81a461ca"     
    }
  }' \
  http://localhost:6013/preview.png  
```

PR #58, opened 2017-12-08
Img preview fix

leohemsted:
Solution is to:

#### pin weasyprint to 0.40  …

We were seeing issues with weasyprint==0.41 where PNG previews of PDFs
sometimes did not render images - in particular, we saw this with logos
disappearing sometimes. The trigger seemed arbitrary, but was often
related to high code points (eg typographical characters, emoji), or
large template contents. Pinning to 0.40 has resolved this.

weasyprint would print out the following warning:

```
   **** Error reading a content stream. The page may be incomplete.
   **** Warning: Pattern stream has unbalanced q/Q operators (too many q's)

   **** This file had errors that were repaired or ignored.
   **** The file was produced by:
   **** >>>> WeasyPrint 0.41 (http://weasyprint.org/) <<<<
   **** Please notify the author of the software that produced this
   **** file that it does not conform to Adobe's published PDF
   **** specification.
```


leohemsted:
update: weasyprint 0.41 works fine locally - but not in docker. Must be something to do with the binary dependencies we install in dockerfile.

I'll investigate, but lets ship this for now

PR #60, opened 2017-12-13
Add page count to header for print.pdf endpoint

kentsanggds:
## What

In order to calculate letter billing we need a page count from the pdf.

## How to review

Using curl we should be able to see the headers generated when calling the print.pdf endpoint

```
curl -v -O \
  -H "Authorization: Token my-secret-key" \
  -H "Content-type: application/json" \
  -d '{
    "template":{
      "subject": "foo",
      "content": "test"
    },
    "values": null,
    "letter_contact_block": "name,\naddress_line_1,\naddress_line_2",
    "dvla_org_id": "001"
  }' \
  http://localhost:6013/print.pdf
```

## Reference

Part of step 5/5 - https://www.pivotaltracker.com/story/show/151511579

### DIFF
--- a/app/preview.py
+++ b/app/preview.py
@@ -45,11 +45,21 @@ def get_logo_filename(dvla_org_id):
         abort(400)
 
 
+def get_page_count(pdf_data):
+    with Image(blob=pdf_data) as image:
+        return len(image.sequence)
+
+
 @preview_blueprint.route("/preview.json", methods=['POST'])
 @auth.login_required
 def page_count():
-    with Image(blob=view_letter_template(filetype='pdf').get_data()) as image:
-        return jsonify({'count': len(image.sequence)})
+    return jsonify(
+        {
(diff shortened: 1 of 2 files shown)


app/preview.py
```diff
@@ -135,12 +135,16 @@ def print_letter_template():
...
                 pdf_data.write(color_mapping(line))
 
-        return send_file(
+        response = send_file(
             BytesIO(pdf_data.result),
             as_attachment=True,
             attachment_filename='print.pdf'
         )
 
+        with Image(blob=pdf_data.result) as image:
```
quis: Can you refactor into a helper method so that the same code isn’t repeated from here: https://github.com/alphagov/notifications-template-preview/blob/8cfae14bc951c72aaf5f9b3e35762df057d7002f/app/preview.py#L49

This will mean that the same tests that cover the page count endpoint will also give us some coverage here.

PR #61, opened 2017-12-22
Added statsd logging to template preview

richardc0:
It is suspected that the template preview app performance is slow.
Added statsd to template preview so that the performance of the
methods can be monitored overtime and so that future updates
can be proved to have a positive impact.

- Added statsd decorator to all preview.py methods
- Added configuration for statsd in init
- Added test to ensure that statsd config is being imported
- Added environment variable to make file to not turn on statds locally
- Added a variable to the manifest file so that statsd is turned on when deployed


venusbb:
Is it possible to test statd locally?

leohemsted:
@venusbb not the pushing of data, but the logs themselves should always print

PR #66, opened 2018-01-08
Make logos in printed letters CMYK

quis:
Letters will look best printed with a CMYK, vector logo. Previews will still look best by using an RGB, raster version of the logo.

We have good-looking RGB logos already using PNGs.

This commit adds the option to specify a version of the logo for printing. Experimentation has found that a CMYK SVG works well, and has its colour space preserved in the outputted PDF.

This commit implements the CMYK version of the HM Government logo only, but refactors the code to make it easy to add more CMYK logos in the future.

Eventually this should be a combination of things stored in the database and logo images stored in S3, so that we can manage logos via platform admin. For expediency we’re going to continue managing these manually for now.

***

hm-cmyk

***

The yellow edge is a bug where the colour replacement algorithm is too greedy. It should be rendering a CMYK blue instead. However fixing it is out of scope for this PR.


static/images/letter-template/hm-government-cmyk.svg
```diff
@@ -0,0 +1,423 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 21.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 85.9 29.2" style="enable-background:new 0 0 85.9 29.2;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:#407CCA;}
```
quis: This is RGB. Needs more investigating…

quis: image

quis:
# Use RGB vector logos for print, then convert

Turns out:
- SVG is the only vector format that works in web pages (which is what our letters are generated from)
- SVG does not support CMYK except in version 2, and there are no  graphics tools which support SVG version 2

So this commit changes things a bit. It keeps:
- support for a new, SVG, version of the logo for printing

It removes:
- the implication that this vector version is CMYK

Luckily, the CMYK conversion stuff we wrote before is able to target the colours in the vector logo. So we just need to make sure we have the colour conversion set up properly to properly map any colours that we have in logos. This is hard coded for now, but we can probably make it more intelligent in the future.

---

image

quis:
# Use common formats for specifying colour values

It’s unusual to specify colour values as numbers from 0.0–1.0. This means manually converting to this format when adding new rows to the colour mapping file.

Most graphics software uses values from:
- 0–255 for RGB triplets
- 0–100% for CMYK quadruplets

By programmatically converting from these formats to the decimal values needed for the transformation we reduce the risk of introducing error by manual conversion, and make it quicker to add new values to the mapping.

quis:
# Add single identity colours to mapping

There’s a fair chance that we’ll one day be sending letters with logos of most government departments. So we should make sure that we get their colours accurate.

The branding for most central government organisations is defined by the ‘single identity’. This commit adds those colours to the mapping. There are other, non single identity, logos which we will also have to deal with later, either by adding to the mapping or writing some code
which takes a best guess at the transformation.

The colour values added here are eyedropped from the document ‘Government Identity System CMYK.pdf’ which can be found in the ‘__AssetLibrary-Design’ folder on Google Drive.

quis:
# Add SVG versions of current single identity logos

We need SVG (ie vector) versions of these logos for printing letters. This commit adds SVG versions of all the single identity logos that we currently print on letters.

There is more work to come to add vector versions of the non-single-identity logos.

---

_Left: PNG, right: SVG_

image
image
image
image
image
image

quis:
# Change GEO from purple to magenta

At some point since the single identity was created in 2012 the Government Equalities Office changed their identity from the same purple as Home Office to the same magenta as DCMS.

This commit updates the logo and colour mapping to match what is:
- in the raster version of the logo
- on GOV.UK: https://www.gov.uk/government/organisations/government-equalities-office

---

_Left: PNG; right: SVG_

image

app/__init__.py
```diff
@@ -4,25 +4,49 @@
...
-    '001': 'hm-government.png',
-    '002': 'opg.png',
-    '003': 'dwp.png',
-    '004': 'geo.png',
-    '005': 'ch.png',
-    '006': 'dwp-welsh.png',
-    '007': 'dept-for-communities.png',
-    '008': 'mmo.png',
-    '500': 'hm-land-registry.png',
+LOGOS = {
```
richardc0: Was there a reason to keep the 001, 002 scheme? It would make more sense to label them as HM_GOV, OPG etc

tests/test_preview.py
```diff
@@ -216,16 +216,19 @@ def test_print_letter_returns_200(print_letter_template):
...
-        os.path.join(
-            os.path.dirname(os.path.realpath(__file__)),
-            '..',
-            'static', 'images', 'letter-template',
-            filename
+    assert get_logo(dvla_org_id).raster == expected_filename
+
+
+@pytest.mark.parametrize('logo', LOGOS.values())
+def test_that_logo_files_exist(logo):
```
richardc0: Add a test to handle the cases where the LOGO has neither a rastor or a vector

app/__init__.py
quis: We need to stay compatible with what’s stored in the database for now. Agree that we should revisit this once we’re not dependent on DVLA storing copies of the logos (which they identify with the 001, 002 scheme.

tests/test_preview.py
quis: Tests added in d92fd96

PR #72, opened 2018-01-10
It is suspected that the template preview app performance is slow.

richardc0:
Added statsd to template preview so that the performance of the
methods can be monitored overtime and so that future updates
can be proved to have a positive impact.

- Added statsd decorator to all preview.py methods
- Added configuration for statsd in init
- Added environment variable to make file to not turn on statds
locally
- Added a variable to the manifest file so that statsd is turned on
when deployed
- Added the statsd prefix for all environments

image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -49,7 +49,7 @@ LOGOS = {
 def load_config(application):
     application.config['API_KEY'] = os.environ['TEMPLATE_PREVIEW_API_KEY']
     application.config['LOGOS'] = LOGOS
-    application.config['NOTIFY_ENVIRONMENT'] = os.environ['NOTIFY_ENVIRONMENT']
+    application.config['NOTIFY_ENVIRONMENT'] = os.environ['NOTIFY_QUEUE_PREFIX']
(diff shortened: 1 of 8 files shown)


manifest-base.yml
```diff
@@ -4,12 +4,14 @@ command: ./scripts/run_app.sh $PORT
 instances: 1
 memory: 2G
 env:
-  NOTIFY_APP_NAME: notify-template-preview
+  NOTIFY_APP_NAME: rc-notify-template-preview
```
venusbb: Need to change this back.

app/__init__.py
```diff
@@ -47,6 +49,16 @@
 def load_config(application):
     application.config['API_KEY'] = os.environ['TEMPLATE_PREVIEW_API_KEY']
     application.config['LOGOS'] = LOGOS
+    application.config['NOTIFY_ENVIRONMENT'] = os.environ['NOTIFY_ENVIRONMENT']
```
venusbb: It would be better to use the name and space_name under VCAP_APPLICATION, similar to here:-
https://github.com/alphagov/notifications-api/blob/master/app/cloudfoundry_config.py#L20

PR #75, opened 2018-01-11
convert pdf to CMYK using ghostscript sSourceObjectICC

leohemsted:
Get rid of the manual editing of PDF files for colour conversion.

Generate the rgb_to_cmyk.icc file by combining the ghostscript default_rgb.icc and default_cmyk.icc (in /usr/share/ghostscript) using the collink command with -f: http://www.argyllcms.com/doc/collink.html#f 
This causes the CMYK output to be K (black) only for exact greyscale colours.

To load that rgb_to_cmyk.icc file we need to pass the control.txt file in to ghostscript as sSourceObjectICC. This transforms any R=G=B into strictly K, which should avoid the grey fuzzy text we've seen with the existing DVLA flow. Any other images go through the normal RGB to CMYK conversion.

https://stackoverflow.com/questions/35596346/ghostscript-convert-pdf-into-cmyk-preserving-pure-black-for-text

@quis


quis:
This works, we should ship it.

The one thing that is going to be trickier to get right is where logos have CMYK colours which are out of gamut in RGB (because we have to store logos in RGB). An example of this is the Department of Energy and Climate Change which is `RGB 0, 159, 227` but `CMYK 100, 0, 0, 0` (ie much more saturated in print that can be displayed on screen):

image

PR #77, opened 2018-01-12
Add remaining vector logos for printing

quis:
This commit adds:
- vector logos for the non-single-identity orgs
- vector and raster logos for a new organisation (Environment Agency)


quis:
# Added DWP bilingual logo

image

quis:
# Remove ability to omit vector logos

We have vector versions of all the logos now. There isn’t a future case where we should be adding new logos without both vector and raster versions.

PR #78, opened 2018-01-12
Added Redis Caching of pdf dataReverted Changes to status.py 

richardc0:
This PR adds in Redis caching of PDF data to improve performance on a new endpoint, this stops the same pdf data being created for subsequent png calls.

A new unique name is generated based on if it is with or without personalisation data. This has a benefit as template preview creates the pdf and created the same data for each page of the pdf in png format, meaning this method can be called up to four times per user view, three of which would recreate the pdf data which takes around 300ms, so a total saving of approx 900ms (Nb. the whole process currently takes 3-4 secs)

- Added redis client to the app
- Added a new endpoint to the preview.py file to use the redis client to cache (if available)
- Added test for the new endpoint


app/preview.py
```diff
@@ -65,6 +65,27 @@ def page_count():
...
+
+    unique_name_dict = {
+        'template_id': json['template']['id'],
+        'version': json['template']['version'],
+        'dvla_org_id': json['dvla_org_id'],
+        'letter_contact_block': json['letter_contact_block'],
+        'values': None if not json['values'] else print_dict(json['values'])
+    }
+
+    return print_dict(unique_name_dict)
```
leohemsted: this currently returns a list - should we cast this to a string at this point?

requirements.txt
```diff
@@ -12,6 +12,7 @@ wand==0.4.4
 jsonschema==2.6.0
 numpy==1.13.3
 PyYAML==3.12
+mock==2.0.0
```
leohemsted: don't need this - mock is in `unittest` in python 2.7 and above

tests/test_preview.py
```diff
@@ -1,16 +1,19 @@
 import json
 import os
-from unittest.mock import patch, Mock
+import uuid
 
 from flask import url_for
 from functools import partial
 import pytest
+from mock import patch, Mock
```
leohemsted: keep this as `from unittest.mock`

tests/test_preview.py
```diff
@@ -250,3 +257,116 @@ def test_that_logo_files_exist(logo):
...
+        client.post(
+            url_for('preview_blueprint.view_letter_template', filetype=filetype, page=page_number),
+            data=notification_data,
+            headers={
+                'Content-type': 'application/json',
+                **auth_header
+            }
+        )
+
+        assert mocked_redis_get.call_count == 1
```
leohemsted: this line is redundant as `assert_called_once_with` moans if it's called 0 or 2 or more times (and elsewhere where it is used)

tests/test_preview.py
```diff
@@ -250,3 +257,116 @@ def test_that_logo_files_exist(logo):
...
+            'dvla_org_id': '001',
+            'letter_contact_block': '123',
+            'values': None
+        }
+
+        assert mocked_redis_get.call_count == 1
+        mocked_redis_get.assert_called_once_with(sorted(unique_name_dict.items()))
+        assert mocked_redis_set.call_count == 1
+
+        mocked_redis_get = mocker.patch('app.preview.current_app.redis_store.get', return_value=response.data)
```
leohemsted: these should be separate unit tests

tests/test_preview.py
```diff
@@ -250,3 +257,116 @@ def test_that_logo_files_exist(logo):
...
+            data=notification_data,
+            headers={
+                'Content-type': 'application/json',
+                **auth_header
+            }
+        )
+
+        assert response.status_code == 400
+        assert mocked_redis_get.call_count == 0
+        assert mocked_redis_set.call_count == 0
```
leohemsted: duplicated line

tests/test_preview.py
```diff
@@ -250,3 +257,116 @@ def test_that_logo_files_exist(logo):
...
+        response = client.post(
+            url_for('preview_blueprint.view_letter_template', filetype='pdf', page='10'),
+            data=notification_data,
+            headers={
+                'Content-type': 'application/json',
+                **auth_header
+            }
+        )
+
+        assert response.status_code == 400
```
leohemsted: i don't follow - what is this test checking? what is this 400 for?

tests/test_preview.py
```diff
@@ -250,3 +257,116 @@ def test_that_logo_files_exist(logo):
 def test_that_logos_must_have_at_least_one_file(partially_initialised_class):
     with pytest.raises(TypeError):
         partially_initialised_class()
+
+
+@pytest.mark.parametrize('filetype, sentence_count, page_number, expected_response_code', [
+    ('png', 10, 1, 200),
+    ('png', 10, 2, 400),
+    ('png', 50, 2, 200),
+    ('png', 50, 3, 400),
```
leohemsted: why do we care about 400 responses here? surely this test should just be "make sure that if there is a value in the cache with the correct key, that value is returned and we don't generate a pdf"

tests/test_preview.py
richardc0: Ones is get one is set

tests/test_preview.py
richardc0: You are right I copied the whole test for cached but the 400s are already tested and the fact caching is turned on makes no difference to the flow. Removed Test.

tests/test_preview.py
richardc0: Removed

tests/test_preview.py
richardc0: You are right I copied the whole test for cached but the 400s are already tested and the fact caching is turned on makes no difference to the flow. Removed Test.

tests/test_preview.py
richardc0: Done

requirements.txt
richardc0: Done

Makefile
```diff
@@ -98,8 +98,10 @@ define run_docker_container
 		-e CI_BUILD_URL=${BUILD_URL} \
 		-e TEMPLATE_PREVIEW_API_KEY=${TEMPLATE_PREVIEW_API_KEY} \
 		-e STATSD_ENABLED= \
-		-e STATSD_PREFIX="{CF_SPACE}" \
+		-e STATSD_PREFIX="dev" \
```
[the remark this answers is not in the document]
richardc0: Changed back

tests/test_preview.py
richardc0: No it should have been testing that a cached pdf is used, but it wasn't.
I've updated the parameters and the code to o that and changed the name of the method to make it clearer.

app/preview.py
richardc0: Not changed. Is there a benefit to doing that?

leohemsted: hmm, if redis is using it as a string then just clarity i guess. it's probably fine here i spose

tests/test_preview.py
```diff
@@ -250,3 +257,71 @@ def test_that_logo_files_exist(logo):
...
+            'version': 1,
+            'dvla_org_id': '001',
+            'letter_contact_block': '123',
+            'values': None
+        }
+
+        mocked_redis_get.assert_called_once_with(sorted(unique_name_dict.items()))
+        assert mocked_redis_set.call_count == 1
+
+        mocked_redis_get = mocker.patch('app.preview.current_app.redis_store.get', return_value=response.data)
```
leohemsted: i'm still not sure why this is one test - let's run through this together when you're free

app/preview.py
servingUpAces: Would it work if we use sorted(d.values()) rather than items?

app/preview.py
```diff
@@ -65,6 +65,27 @@ def page_count():
     )
 
 
+@statsd(namespace="template_preview")
+def get_pdf_redis_key(json):
+    def print_dict(d):
+        """
+        From any environment/system will return the same string representation of a given dict.
+
+        ASSUMPTION: all dict values are immutable (eg int, none, str)
```
quis: This assumption is incorrect. Personalisation values can be lists.

leohemsted: ahhhh nuts 😓

app/preview.py
leohemsted: if someone uses the same placeholders in a different order while previewing it'd not be able to distinguish between them - so you could get the wrong PDF returned

app/preview.py
quis: Just needs a slightly smarter algorithm here to flatten everything into something hashable. Not a massive change I don’t think.

leohemsted: so the implementation's changed (we tried using `frozenset`) and the comment's out of date - immutability is no longer a requirement, but sortability is - you can't sort a list of disparate types, so i think we'll just cast everything to a string before sorting.

leohemsted: want to avoid hashing, at least to start, so we can read the redis keys for debugging.

leohemsted: we've added a test that makes sure it does sort mixed value types (`str`, `int`, `None`, `bool`, and `list`)

PR #79, opened 2018-01-17
remove numpy

leohemsted:
We don't use it any more since the new cmyk conversion strategy


requirements.txt
```diff
@@ -1,16 +1,10 @@
-# WARNING
```
quis: Gonna assume you know what you’re doing removing this.

leohemsted: yeah, dockerfile was updated to automatically read the requirements.txt and install it

PR #80, opened 2018-01-17
Update ghostscript to produce 1.6 PDFs rather than 1.5

kentsanggds:
## What 

DVLA require PDF versions 1.6, this PR will update the version to 1.6 from 1.5 and also use utils 23.4.3 to remove the colon from the notify-tag.


quis:
image

👍

PR #83, opened 2018-01-18
Add -q to only output pdf data and not standard output

kentsanggds:
## What

DVLA were not able to process the PDFs because they did not appear to be valid pdfs. 
It seems that the standard output of the ghostscript was being added to the pdf data stream, adding  `-q` should fix this and make the pdf valid.

### DIFF
--- a/app/transformation.py
+++ b/app/transformation.py
@@ -12,6 +12,7 @@ def convert_pdf_to_cmyk(input_data):
     stdout, _ = subprocess.Popen(
         [
             'gs',
+            '-q',
             '-o',
             '-',
             '-dCompatibilityLevel=1.7',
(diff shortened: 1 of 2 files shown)


quis:
You could write a test that checks if the file starts with `%PDF-1.7` – this would stop this happening again, and stops someone accidentally removing the line that sets the PDF version.

For example, before this change running `head print.pdf` gives me this output:
```pdf
GPL Ghostscript 9.20 (2016-09-26)
Copyright (C) 2016 Artifex Software, Inc.  All rights reserved.
This software comes with NO WARRANTY: see the file PUBLIC for details.
Processing pages 1 through 1.
Page 1
%PDF-1.6
%�쏢
7 0 obj
<</Length 8 0 R/Filter /FlateDecode>>
stream
```

PR #85, opened 2018-01-24
Added commandline parameters to turn off compression for PDF DVLA

richardc0:
processing. Whilst compression can be handled their end it means that
processing the  pdfs takes longer. This speeds up the process.

These options are not documented comprehensively by Ghostscript and were
supplied by DVLA.

This website also mentions that they turn off compression
http://ansuz.sooke.bc.ca/software/ps-pdf-tricks.php

Ghostscript docs here https://www.ghostscript.com/doc/current/Ps2pdf.htm


quis:
Should double check that this doesn’t affect the CMYK conversion of logos. Let me know if you need a hand doing that.

PR #86, opened 2018-01-26
Ensure print version of DWP logo is R:0 G0 B:0

quis:
Parts of it had a couple percent for the RGB values – means it doesn’t convert to pure CMYK black.

---

image


static/images/letter-template/dwp.svg
```diff
@@ -3,238 +3,225 @@
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
 	 viewBox="0 0 989.6 822.1" style="enable-background:new 0 0 989.6 822.1;" xml:space="preserve">
 <style type="text/css">
-	.st0{fill:#010202;}
-	.st1{fill:none;stroke:#00B0AD;stroke-width:15;}
-	.st2{fill:#020203;}
+	.st0{fill:none;stroke:#00B0AD;stroke-width:15;}
```
quis: `#00B0AD` is the DWP green. All other colours are unspecified so default to black.
image

PR #98, opened 2018-02-23
Add precompiled-preview.png endpoint to extract a png from a pdf

kentsanggds:
## What

POSTing to the endpoint `precompiled-preview.png` with a base64 encoded pdf payload and a request argument page indicating the page number will return a PNG of the page requested. 

If no page number is given it will default to the first page

## Reference 

https://www.pivotaltracker.com/story/show/155323846


app/preview.py
```diff
@@ -139,6 +151,32 @@ def view_letter_template(filetype):
...
+@auth.login_required
+@statsd(namespace="template_preview")
+def view_precompiled_letter():
+    try:
+        encoded_string = request.get_data()
+
+        if not encoded_string:
+            abort(400)
+
+        pdf = base64.decodestring(encoded_string)
```
richardc0: Do we trust the sender to give us a PDF? or should we do a check here?

app/preview.py
```diff
@@ -139,6 +151,32 @@ def view_letter_template(filetype):
...
+
+        return send_file(**png_from_pdf(
+            pdf, page_number=int(request.args.get('page', 1)), hide_notify=True
+        ))
+
+    # catch invalid pdfs
+    except MissingDelegateError as e:
+        current_app.logger.error(str(e))
```
richardc0: Add some context here? What fails when there is a a MissingDelegateError?
 e.g. current_app.logger.error("Failed to decode base64String" + str(e))

PR #102, opened 2018-03-06
Make it more consistent with the API

sakisv:
Copy things from notifications-api to ensure that we're starting
properly and we're shipping logs to cloudwatch

----

Need to double check that we're exporting logs in the correct format and make sure we test it manually on preview before it gets merged.

### DIFF
--- a/app/gunicorn_config.py
+++ b/app/gunicorn_config.py
@@ -1,0 +1,25 @@
+import os
+import sys
+import traceback
+
+workers = 10
+errorlog = "/home/vcap/logs/gunicorn_error.log"
+bind = "0.0.0.0:{}".format(os.getenv("PORT"))
+
+
+def on_starting(server):
+    server.log.info("Starting Notifications template preview")
+
+
+def worker_abort(worker):
+    worker.log.info("worker received ABORT {}".format(worker.pid))
+    for threadId, stack in sys._current_frames().items():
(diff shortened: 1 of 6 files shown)


leohemsted:
we're definitely not exporting logs in the right format

leohemsted:
app/__init__.py should probably say something like

```
from notifications_utils import logging

def create_app():
   [...]
    logging.init_app(application, statsd_client)
```

at line 68-70

requirements.txt
```diff
@@ -16,5 +16,5 @@ git+https://github.com/alphagov/notifications-utils.git@23.7.0#egg=notifications
 
 # PaaS requirements
 gunicorn==19.7.1
-# awscli>=1.11,<1.12
-# awscli-cwlogs>=1.4,<1.5
+awscli>=1.11,<1.12
```
leohemsted: we don't pin `awscli` in the other repos 🤔

app/gunicorn_config.py
```diff
@@ -0,0 +1,25 @@
+import os
+import sys
+import traceback
+
+workers = 10
+errorlog = "/home/vcap/logs/gunicorn_error.log"
```
leohemsted: do we want to move the dockerfile working directory so that the app is in `/home/vcap/app`?

sakisv: I don't see why not 👍

PR #103, opened 2018-03-14
Make sure directories exist before touching files

sakisv:
This runs before the application starts and if the directories are not
there it fails and doesn't allow the application to start.

### DIFF
--- a/scripts/run_app_paas.sh
+++ b/scripts/run_app_paas.sh
@@ -17,6 +17,7 @@ function check_params {
 
 function configure_aws_logs {
   # create files so that aws logs agent doesn't complain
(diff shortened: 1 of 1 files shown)


scripts/run_app_paas.sh
```diff
@@ -17,6 +17,7 @@ function check_params {
 
 function configure_aws_logs {
   # create files so that aws logs agent doesn't complain
+  mkdir -p /home/vcap/logs/
```
leohemsted: if the folder already exists won't this explode

sakisv: No

leohemsted: okay

PR #109, opened 2018-03-29
Fix cmyk preview precompiled

kentsanggds:
## What

CMYK pdfs were showing as inverted due to the way colour is handled in the CMYK colour space vs how it is handled in the RGB colour space. 
To fix this we transform the colour space to CMYK to match the PDF.

## How to review

Upload a CMYK pdf via API and view on the Admin site, you should be able to view the colours without them being inverted.


quis:
Screenshots!

PR #125, opened 2018-05-11
Bump utils for new newline rendering

quis:
Brings in:
- [x] https://github.com/alphagov/notifications-utils/pull/476


quis:
Closing this to do page breaks properly.

PR #129, opened 2018-05-15
Generate letters from HTML, not data

quis:
This is a bit of a mega commit 😳 

It rearchitects the way template preview app works to take a string of HTML as input, rather than data. In other words, it takes away the building-up-the-letter part of the work away from this app.

This means that:
- it’s easier to work on the letter template locally, because you can make local changes without having to modify this app (which runs in Docker, doesn’t hot load, can’t use local utils, etc)
- it’s completely deterministic, ie same input always gives same output…
- …which means that we can cache way more aggressively


quis:
We should also look into whether S3 would be a better place to cache these files.

PR #130, opened 2018-05-16
Add logo for Tyne and Wear Fire and Rescue service

quis:
This is a simplified version of their actual logo, with a lot of the gradients and strokes removed because:

- they won’t show up at the small size
- they made the file too complicated for our PDF thing to handle


quis:
# Also add Thames Valley Police logo

image

PR #151, opened 2018-06-21
Added functions to validate a PDF against the letter template

richardc0:
Added functions to validate a PDF against the letter template. This might be used by admin to validate the template through a tool for new services and also as a way to check each PDF as they work their way through the system. This may need to be done asynchronously as the synchronous call might be too slow in a similar way to the anti-virus scan.

* Added the validate and overlay endpoints to precompiled.py
* Updated the Dockerfile with the new dependencies
* Added tests to test the validation functions
* Removed superfluous tasks from the Makefile
* Update the requirements required by the validation methods

This is a branch from the rc_add_tag_to_precompiled so only the last commit needs review as the rest in in this PR https://github.com/alphagov/notifications-template-preview/pull/149 (that is waiting on the test from DVLA to pass).


app/precompiled.py
```diff
@@ -0,0 +1,340 @@
...
+        abort(400)
+
+    # catch invalid pdfs
+    except PdfReadError as e:
+        current_app.logger.warn("Failed to read PDF", str(e))
+        abort(400)
+
+    except Exception as e:
+        current_app.logger.error(str(e))
+        raise e
```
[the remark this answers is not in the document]
leohemsted: in fact, how do you feel about adding error handlers for those exceptions to this blueprint? Something like:

```python
@precompiled_blueprint.error_handler(binascii.Error)
def handle_ base64_error(e):
    msg = "Unable to decode the PDF data: {}".format(e)
    current_app.logger.warn(msg)
    return jsonify(message=msg), 400
```

app/precompiled.py
```diff
@@ -0,0 +1,340 @@
...
+    packet.seek(0)
+    new_pdf = PyPDF2.PdfFileReader(packet)
+
+    page.mergePage(new_pdf.getPage(0))
+    output.addPage(page)
+
+    page_num = 1
+
+    # For each subsequent page its just the body of text
+    while page_num < pdf.numPages:
```
leohemsted: `for page_num in range(1, pdf.numPages):`

app/precompiled.py
```diff
@@ -0,0 +1,340 @@
...
+    new_pdf = PdfFileReader(packet)
+
+    new_page = new_pdf.getPage(0)
+    new_page.mergePage(page)
+    output.addPage(new_page)
+
+    page_num = 1
+
+    # add the rest of the document to the new doc. NOTIFY only appears on the first page
+    while page_num < pdf.numPages:
```
leohemsted: `for page_num in range(1, pdf.numPages):`

app/precompiled.py
```diff
@@ -0,0 +1,340 @@
...
+    Checks each pixel of the image to determine the colour - if any pixel is not white return false
+    :param PyPDF2.PdfFileReader src_pdf: PDF from which to take pages.
+    :return: False if there is any colour but white, otherwise true
+    """
+
+    dst_pdf = PyPDF2.PdfFileWriter()
+
+    page_num = 0
+
+    while page_num < src_pdf.numPages:
```
leohemsted: `for page_num in range(src_pdf.numPages):`

tests/test_precompiled.py
```diff
@@ -0,0 +1,459 @@
...
+        data=blank_page,
+        headers={
+            'Content-type': 'application/json',
+            **auth_header
+        }
+    )
+
+    assert response.status_code == 200
+    json_data = json.loads(response.get_data())
+    assert json_data['result'] is True
```
leohemsted: small thing - since flask 1.0 you can now call `response.json` rather than having json.loads

tests/test_precompiled.py
```diff
@@ -0,0 +1,459 @@
...
+
+    can.mock_canvas = mocker.patch('app.precompiled.canvas.Canvas', return_value=can)
+
+    file_data = base64.b64decode(multi_page_pdf)
+
+    # It fails because we are mocking but by that time the drawString method has been called so just carry on
+    try:
+        add_notify_tag_to_letter(BytesIO(file_data))
+    except Exception:
+        pass
```
leohemsted:
```
with pytest.raises(SpecificException):
    add_notify_tag_to_letter(BytesIO(file_data))
```

i don't want to get stung by tests not testing stuff if we refactor the code underneath

tests/test_precompiled.py
```diff
@@ -0,0 +1,459 @@
...
+    assert pdf_new.getPage(2).extractText() == pdf_original.getPage(2).extractText()
+    assert pdf_new.getPage(3).extractText() == pdf_original.getPage(3).extractText()
+
+
+def test_add_notify_tag_to_letter_correct_margins(mocker):
+    file_data = base64.b64decode(multi_page_pdf)
+    pdf_original = PyPDF2.PdfFileReader(BytesIO(file_data))
+
+    can = Canvas(None)
+    # mock_canvas = mocker.patch.object(can, 'drawString')
```
leohemsted: i think we can remove this commented out line

tests/test_precompiled.py
```diff
@@ -0,0 +1,459 @@
...
+
+    x = mm_from_left_of_page * mm
+
+    # page.mediaBox[3] Media box is an array with the four corners of the page
+    # We want height so can use that co-ordinate which is located in [3]
+    # The lets take away the margin and the ont size
+    y = float(pdf_original.getPage(0).mediaBox[3]) - (float(mm_from_top_of_the_page * mm + 6 - 1.75))
+
+    can.drawString.assert_called_once()
+    can.drawString.assert_called_with(x, y, "NOTIFY")
```
leohemsted: can combine these two lines into `assert_called_once_with`

PR #162, opened 2018-07-03
Added endpoint for Precompiled CMYK Conversion

richardc0:
It is not guaranteed that a PDF will be CMYK when received from a service, hecne this endpoint converts the PDF to CYMK. This is so the service does not have to worry about the CYMK conversion and ensures the PDF version is always consistent for the DVLA. 
This may end up being part of an async tasks which converts and adds the NOTIFY tag and validates etc.

Note. Its tricky to validate that the colour space has changed even using conventional tools so validate that the PDF version has changed and as such we know the Ghostscript process has run.

* Added new endpoint
* Added a generic error handler to handle common errors so the the preview.py code is more readable
* Added tests for the new endpoint
* Added test coverage display on test completion.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,9 +3,11 @@ import os
 from contextlib import suppress
 from hashlib import sha1
 
+import PyPDF2
+
 from app.transformation import Logo
 
-from flask import Flask
+from flask import Flask, jsonify
 from flask_httpauth import HTTPTokenAuth
 
 from notifications_utils import logging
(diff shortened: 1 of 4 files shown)


app/__init__.py
```diff
@@ -134,3 +138,26 @@ def new_function():
...
+    @app.errorhandler(Exception)
+    def exception(error):
+        app.logger.exception(error)
+
+        if hasattr(error, 'message'):
+            # error.code is set for our exception types.
+            return jsonify(result='error', message=error.message or ""), error.code or 500
+        elif hasattr(error, 'code'):
+            # error.code is set for our exception types.
+            return jsonify(result='error'), error.code or 500
```
servingUpAces: Is there something else in the exception that we can add as the message?

PR #177, opened 2018-08-07
Precompiled letter sanitising - notify tag and address block checking

leohemsted:
Adds a new `/precompiled/sanitise` endpoint.

Accepts: A base64 encoded PDF file (todo: do we want to change this? This should probably just be a raw file)
Returns: A regular pdf file.

What it does:

* Identify if notify tag is present
  - if not, adds a notify tag to the first page
* validates the boundaries aren't exceeded
  - if they are, logs an error. But doesn't return 400, because I suspect most letters we get won't pass this.
* Extracts the address from the first page
* puts a white block over the address
* re-writes the address, in arial 8pt. 
* returns the new, improved, all-singing all-dancing PDF.

Suggestions on how to refactor `precompiled.py` so that it doesn't start with a bazillion random variables are welcome


leohemsted:
to test (from API):

```python
import requests
resp = requests.post(
    'http://localhost:6013/precompiled/sanitise',
    data=base64_encoded_pdf,
    headers={'Authorization': 'Token my-secret-key'}
)
resp.raise_for_status()
with open('output.pdf'), 'wb') as f:
    for chunk in resp.iter_content(chunk_size=128):
        f.write(chunk)
```

PR #178, opened 2018-08-10
Bump disk quota to 2G

sakisv:
Even newly created instances got their disk to 850MB out of 1GB.

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -4,6 +4,7 @@ applications:
 
   instances: 1
   memory: 1G
+  disk_quota: 2G
 
   buildpack: python_buildpack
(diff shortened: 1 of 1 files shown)


quis:
History:
- memory up from from 512MB to 2GB:https://github.com/alphagov/notifications-template-preview/commit/e9b209a022af9f42f774f79b414dbf63c1c8c67a#diff-336218f780d169ff24b1103188ca554fR8
- memory reduced from 2GB to 1GB: https://github.com/alphagov/notifications-template-preview/commit/61eb51e3a98e603c79d83541fb60c0d02b39d571#diff-336218f780d169ff24b1103188ca554fR4

PR #207, opened 2018-09-13
allow content on the right hand side of precompiled pdfs

leohemsted:
we have already succesfully processed precompiled pdfs with content on
the right - turns out DVLA don't care about it as much as they care
about the other margins. This commit removes the right hand border.

It also does some refactoring to make the pink overlay.png and the
validation endpoint more similar, and renames some variable names for clarity

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -213,16 +213,17 @@ def _add_no_print_areas(src_pdf, overlay=False):
     packet = BytesIO()
     can = canvas.Canvas(packet, pagesize=A4)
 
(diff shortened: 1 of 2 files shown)


quis:
Can we ask what the actual limit for the right-hand edge is please? Because I bet it’s not 0, and we don’t wanna have to go through this migration again when we find that out…

yahoopete:
The printable margin is effectively right to the edge (confirmed by DVLA) so suggest we go ahead with this now.

PR #213, opened 2018-09-25
Precompiled validate with preview

CrystalPea:
For the new admin endpoint to validate boundaries of pre-compiled
letter PDFs we need to display all PNG pages and any validation
errors on one page. In order to avoid storing the PDF and making
multiple separate requests to template-preview we add a new flag
to the validate route that returns base64-encoded PNG images of
all PDF pages that could be included into the admin app page using
data-uris.

As a part of this we:
- added a preview flag to validate precompiled endpoint
- with flag, it returns preview pngs, with overlay if
validation failed, and without overlay otherwise
- upgraded overlay_template_areas to handle multiple pages pdf
- created pdf_to_pngs function that converts multiple-pages
pdf to pngs and returns a list of png pages

Pivotal ticket: https://www.pivotaltracker.com/story/show/159697071

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -124,7 +124,7 @@ def validate_pdf_document():
     else:
         data['message'] = 'Your PDF passed the layout check'
         file_data = rewrite_address_block(BytesIO(encoded_string))
-        pages = pngs_from_pdf(file_data, overlay=True)
+        pages = pngs_from_pdf(file_data)
 
     data['pages'] = [
         base64.b64encode(page.read()).decode('ascii') for page in pages
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -105,7 +105,7 @@ def add_tag_to_precompiled_letter():
 @statsd(namespace="template_preview")
 def validate_pdf_document():
     encoded_string = request.get_data()
-    generate_preview_pngs = request.args.get('include_preview') in ['true', '1']
+    generate_preview_pngs = request.args.get('include_preview') in ['true', 'True', '1']
 
     if not encoded_string:
         abort(400)
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -1,9 +1,10 @@
+import base64
 import subprocess
 from io import BytesIO
 
 from PIL import ImageFont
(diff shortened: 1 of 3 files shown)


app/precompiled.py
```diff
@@ -104,17 +105,32 @@ def add_tag_to_precompiled_letter():
...
+
+    if not generate_preview_pngs:
+        return jsonify(data)
+
+    if not data['result']:
+        data['message'] = 'PDF contents outside printable areas'
+        pages = overlay_template_areas(BytesIO(encoded_string), overlay=True)
+
+    else:
+        data['message'] = 'OK'
```
CrystalPea: @karlchillmaid Hi Karl, do you have a better idea for an OK message? It will appear on a banner on the page where a user can validate if their pdf document is within printable boundaries.

We set the validation failed message to 'PDF contents outside printable areas', but we are also opened to change ☺️

karlchillmaid: Validation fail: `Content in this PDF is outside the printable area`

I'm working on a more descriptive confirmation message than 'OK', but I'll need some more information about the user journey first.

karlchillmaid: @CrystalPea I'd like to try `Your PDF passed the layout check` for the time being.

We can revisit it as and when this feature develops or becomes self-service.

app/preview.py
```diff
@@ -32,25 +32,42 @@ def hide_notify_tag(image):
...
     with Image(width=pdf_width, height=pdf_height) as image:
 
         if pdf_colorspace == 'cmyk':
             image.transform_colorspace('cmyk')
 
-        image.composite(page, top=0, left=0)
+        image.composite(pdf_page, top=0, left=0)
         if hide_notify:
             hide_notify_tag(image)
         converted = image.convert('png')
```
allait: This isn't related to our changes, but I think this should be `with image.convert('png') as converted` since that's what most `wand` examples are using. As far as I understand it might help avoid memory leaks if something fails during the conversion.

CrystalPea: Ahh, cool, well-spotted 👍 . I will change it ☺️

PR #219, opened 2018-10-10
pdf validation now tells you which pages the errors were on

leohemsted:
no reason why we can't show this data to people trying to validate their pdf. Replaced the validation function which used to return a True/False boolean, now returns list of pages with errors.

image

also stopped the address boxes overlapping.

image

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -35,7 +35,7 @@ BORDER_FROM_LEFT_OF_PAGE = 15.0
 BORDER_FROM_RIGHT_OF_PAGE = 15.0
 BODY_TOP_FROM_TOP_OF_PAGE = 95.00
 
-SERVICE_ADDRESS_FROM_LEFT_OF_PAGE = 120.0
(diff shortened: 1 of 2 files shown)


app/precompiled.py
```diff
@@ -110,20 +111,30 @@ def validate_pdf_document():
...
+        'result': len(invalid_pages) == 0
     }
 
     if not generate_preview_pngs:
         return jsonify(data)
 
-    if not data['result']:
-        data['message'] = 'Content in this PDF is outside the printable area'
-        pages = overlay_template_areas(BytesIO(encoded_string), overlay=True)
+    file_data = rewrite_address_block(BytesIO(encoded_string))
```
allait: I don't think we should rewrite address in a file that's known to exceed printable area. The address logic relies on the page margins which are likely to be wrong in this case.

leohemsted: after discussion, will revert change

PR #227, opened 2018-10-22
Build latest versions of ghostscript and poppler

allait:
### Use git commit SHA as the Docker image tag

Due to how CloudFoundry stages docker containers deployed apps will scale using the latest container version tagged with a given tag.

This means that Docker tags need to be immutable in order to avoid accidentally starting up instances with a failed release code.

Changing the image tag to use a short git commit SHA (and a `-dirty` suffix for builds with uncommitted changes) makes sure that each release gets it's own separate tag, which makes it easier to rollback and deploy the app locally.

### Move docker image to a Docker Hub organization

govuknotify is a user account that is used by Jenkins to push images to Docker Hub.

Switching images to an organization allows us to give permissions to push containers from dev machines without having to switch to Jenkins user credentials.

Jenkins user account is an owner of the organization, so it should be able to push images without any changes to our CI configuration.

### Switch the order of tests and code style check

Running flake8 before running tests makes it harder to try experimental changes since it requires all code to be properly formatted right away.

### Remove buildpack from PaaS manifest

Buildpacks aren't used for docker deployments to CF. While cf-cli used to ignore them in the past newer versions seem to treat it as an error and fail the release.

### Build latest versions of ghostscript and poppler

Upgrading gs to 9.25 solves the problem with RGB-to-CMYK pre-compiled PDF conversion creating an invalid PDF on some files.

We're running into a number of issues with our PDF/PNG generation related to version of the underlying tools we're using. Relying on system packages makes it hard to upgrade/downgrade these tools to a specific version since there's usually only one available in the system's package repository.

As a partial workaround for this we used to use a mix of Debian Jessie and Stretch repositories to install package versions that were close enough to what we wanted. This approach still left us with few choices of available versions (most of them not up-to-date) and made apt hard to work with (since it now required explicitly choosing the Debian release).

Instead, this switches our base container to the latest stable Debian release (Stretch) and compiles the required versions of the tools we rely on after installing all the build dependencies.

This increases the time it takes to build the container from scratch and can potentially lead to conflicts with system packages (but this hopefully shouldn't be a big problem since not a lot of things rely on pdf libraries), but it gives us direct control over gs/poppler versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,8 @@ APP_VERSION_FILE = app/version.py
 GIT_COMMIT ?= $(shell git rev-parse HEAD 2> /dev/null || cat commit || echo "")
 
 BUILD_TAG ?= notifications-template-preview-manual
-BUILD_NUMBER ?= manual
+BUILD_NUMBER ?= $(shell git describe --always --dirty)
 BUILD_URL ?= manual
-DEPLOY_BUILD_NUMBER ?= ${BUILD_NUMBER}
 
 TEMPLATE_PREVIEW_API_KEY ?= "my-secret-key"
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -29,7 +29,15 @@ $(eval export CF_HOME)
 CF_SPACE ?= sandbox
 
 DOCKER_IMAGE = govuknotifyorg/notifications-template-preview
-DOCKER_IMAGE_TAG = $(shell git describe --always --dirty)
+
+# When promoting a build, use the original job build number
+# Otherwise use either the CI build number or the git commit SHA
+ifdef DEPLOY_BUILD_NUMBER
```
leohemsted: won't this always be defined (from line 12?)

PR #229, opened 2018-10-23
Use letter filename if available instead of dvla_org_id

klssmith:
In the preview_schema, dvla_org_id is now optional instead of required.

For POST endpoints which were expecting to receive the dvla_org_id we
now check for a letter filename in the data first and only use the
dvla_org_id if a filename is not provided. notifications-api and
notifications-admin are already passing the filename through, so this
commit ensures that we now start using this value.

[Pivotal story](https://www.pivotaltracker.com/story/show/159991865)

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -27,9 +27,9 @@ preview_schema = {
             "required": ["subject", "content"]
         },
         "dvla_org_id": {"type": ["string", "null"]},
-        "filename": {"type": ["string", "null"]},
+        "filename": {"type": "string"},
         "date": {"type": ["string", "null"]},
     },
-    "required": ["letter_contact_block", "template", "values"],
+    "required": ["letter_contact_block", "template", "values", "filename"],
     "additionalProperties": False,
 }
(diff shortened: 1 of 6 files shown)


app/schemas.py
```diff
@@ -26,10 +26,10 @@ def get_and_validate_json_from_request(request, schema):
             },
             "required": ["subject", "content"]
         },
-        "dvla_org_id": {"type": "string"},
+        "dvla_org_id": {"type": ["string", "null"]},
         "filename": {"type": ["string", "null"]},
         "date": {"type": ["string", "null"]},
     },
-    "required": ["letter_contact_block", "template", "values", "dvla_org_id"],
+    "required": ["letter_contact_block", "template", "values"],
```
servingUpAces: If filename is always passed in now, should we make it required?

klssmith: 👍 Yes, it makes sense to make `filename` required here instead of in a later PR so that it's not possible to have no value for `filename` and no value for `dvla_org_id` either - I've pushed a commit.

PR #249, opened 2018-12-14
Change the address margin, and log failures

klssmith:
We want to change the address margin for precompiled letters, but before
enforcing the new margin this commit logs any notifications that would
fail the new validation. The 'Letter Validation Preview' page in
notifications-admin has been updated to always use the new margin, since
this is only accessible by platform admin users.

[Pivotal story](https://www.pivotaltracker.com/story/show/162310239)


tests/pdf_consts.py
```diff
@@ -23,3 +23,5 @@ def file(filename):
 rgb_image_pdf = file('tests/test_pdfs/rgb_image.pdf')
 
 cmyk_and_rgb_images_in_one_pdf = file('tests/test_pdfs/cmyk_and_rgb_in_one_pdf.pdf')
+
+new_address_margin = file('tests/test_pdfs/new_address_margin.pdf')
```
allait: It doesn't really matter, but I'd just call it `address_margin`. It's only new right now, but is unlikely to be renamed in the future.

tests/test_precompiled.py
```diff
@@ -91,6 +92,28 @@ def test_precompiled_validation_with_preview_calls_overlay_if_pdf_out_of_bounds(
     assert json_data['pages'] == ['SSdtIGEgcG5n']
 
 
+def test_precompiled_validation_with_preview_calls_overlay_if_content_in_new_address_margin(
```
allait: It sounds like this tests asserts that the mock was called, but I don't think it actually does at the moment

PR #263, opened 2019-01-11
[Spike] Generate PNG versions of logos from SVG

quis:
Currently we do this manually. We can make our process more efficient by automating it.

This also has the benefit of being able to reuse the caching code already in template preview, so we can generate the PNG on the fly then cache it for next time, and we don’t have to write extra code to put the converted file back to S3.

---

image


quis:
I think this will cause the same contention issues we had before, where the app makes a request to itself and locks up. So this exact implementation is probably not right.

Should probably have the admin app use this endpoint to generate the PNG the first time someone uploads a new logo, then save it to the logo bucket.

app/logo.py
```diff
@@ -0,0 +1,41 @@
...
+    svg_file_url = '{}/static/images/letter-template/{}'.format(
+        current_app.config['LETTER_LOGO_URL'],
+        get_logo_from_filename(logo).vector,
+    )
+    try:
+        return send_file(
+            filename_or_fp=_get_png_from_svg(svg_file_url),
+            mimetype='image/png',
+        )
+    except Exception as e:
```
leohemsted: we already have exception handlers in `app/__init__.py` - we don't need it here

tests/test_logo.py
```diff
@@ -0,0 +1,10 @@
+from io import BytesIO
+
+
+def test_logo(mocker, app, client, auth_header):
```
leohemsted: can we get tests for:

* request returns 404
* file is returned but it isn't a valid svg?

app/logo.py
```diff
@@ -0,0 +1,36 @@
...
+from io import BytesIO
+from notifications_utils.statsd_decorators import statsd
+
+from app.preview import get_logo_from_filename
+
+logo_blueprint = Blueprint('logo', __name__)
+
+
+@logo_blueprint.route("/<logo>.svg.png", methods=['GET'])
+@statsd(namespace="template_preview")
```
quis: Should this endpoint require authentication (so only the admin app can talk to it)? I only removed it to make it easier to test locally.

```suggestion
@statsd(namespace="template_preview")
@auth.login_required
```

servingUpAces: sure

PR #315, opened 2019-04-03
Invert overlay to show where validation failed

CrystalPea:
Part of this story: https://www.pivotaltracker.com/story/show/164798738

#### Screenshot:
Screen Shot 2019-04-04 at 11 58 00

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -150,10 +150,10 @@ def validate_pdf_document():
     return jsonify(data)
 
 
-@precompiled_blueprint.route("/precompiled/overlay.png", methods=['POST'])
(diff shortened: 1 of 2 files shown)


app/precompiled.py
```diff
@@ -162,15 +162,26 @@ def overlay_template():
...
-
-    return send_file(
-        filename_or_fp=overlay_template_areas(
-            file_data,
-            int(request.args.get('page', 1)),
-            not validate
-        ),
-        mimetype='image/png',
-    )
+    invert = request.args.get('invert') in ['true', '1']
```
leohemsted: this isn't actually needed as it turns out this function isn't called from any of our other services

leohemsted: also the validate request arg isn't used

PR #338, opened 2019-05-14
zdt deploy

leohemsted:
nb: moved command from manifest to procfile

todo: test this, since it's from a docker container so slightly different to existing process. also confusing NOTIFY_APP_NAME and CF_APP stuff.


leohemsted:
@sakisv it's used in log groups and such (and is different to CF_APP in places eg https://github.com/alphagov/notifications-api/blob/master/manifest.yml.j2#L57) but I agree I think it's unnecessary.

PR #349, opened 2019-07-08
Check page size and orientation

CrystalPea:
Recently one of our users included a landscape-oriented page as a part of their precompiled letter. 
This stopped the printing process at DVLA. 

We now want to check for landscape-oriented pages and fail validation for such pages to avoid such situations in the future.

Ideally we would check if the page is A4 size at the same time. So here we start silently monitoring if letters going through our system are A4 size. If we do not see excessive number of logs about letters failing this check, we will make it live.


app/precompiled.py
```diff
@@ -234,9 +225,54 @@ def overlay_template_areas(src_pdf, page_number=None, overlay=True):
...
 
-def get_invalid_pages(src_pdf):
-    pdf_to_validate = _overlay_printable_areas(src_pdf)
-    return list(_get_out_of_bounds_pages(PdfFileReader(pdf_to_validate)))
+def get_invalid_pages_with_message(src_pdf):
+    message = ""
+    invalid_pages = []
+    are_pages_A4_portrait_oriented, invalid_pages = _are_pages_A4_portrait_oriented(src_pdf)
+    if not are_pages_A4_portrait_oriented:
+        message = 'PDF not conforming to A4 size portrait orientation on '
```
karlchillmaid: Suggested content change: `The page orientation is landscape instead of portrait on `

karlchillmaid: @CrystalPea is the page orientation the only thing that will trigger this message? Or is it checking that the size is A4 as well?

CrystalPea: @karlchillmaid for now only orientation, in the future we also plan to add size checker, but first we want to run it silently for a while to see if our users are within limits without failing them :)

karlchillmaid: OK. If we decide to turn that on, then page size will probably need a unique validation error message too.

app/precompiled.py
```diff
@@ -234,9 +225,55 @@ def overlay_template_areas(src_pdf, page_number=None, overlay=True):
...
+        # page size is in points, there are 72 points in an inch and 25.4 milimeters in an inch,
+        # hence the transformations below to get size in milimeters
+        page_height = float(page.mediaBox.getHeight()) / 72 * 25.4
+        page_width = float(page.mediaBox.getWidth()) / 72 * 25.4
+
+        # are pages A4 portrait oriented. For now we run this silently to check if our users' PDFs mostly comply.
+        # If we decide to fail PDFs that don't meet those conditions, we should probably also allow for
+        # height in range(209, 212) and width in range(296, 299) if page.get('/Rotate') in [90, 270]
+        # (that is a PDF that was originally landscape but has been rotated to be portrait-oriented)
+        if int(page_height) not in range(296, 299) or int(page_width) not in range(209, 212):
```
leohemsted: there's actually https://docs.python.org/3/library/math.html#math.isclose which i found out about just now

that might be a bit neater, because then rather than having to manually add the error range to the actual page size definitions we can use those: `math.isclose(page_height, 297, abs_tol=2)`

thoughts?

app/precompiled.py
```diff
@@ -234,9 +225,55 @@ def overlay_template_areas(src_pdf, page_number=None, overlay=True):
...
+def _are_pages_A4_portrait_oriented(src_pdf):
+    pdf = PdfFileReader(src_pdf)
+    invalid_pages = []
+    for page_num in range(0, pdf.numPages):
+        page = pdf.getPage(page_num)
+
+        # page size is in points, there are 72 points in an inch and 25.4 milimeters in an inch,
+        # hence the transformations below to get size in milimeters
+        page_height = float(page.mediaBox.getHeight()) / 72 * 25.4
+        page_width = float(page.mediaBox.getWidth()) / 72 * 25.4
```
leohemsted: i think it would be good to use the units included in reportlab - `mediabox.getHeight() / mm` should do it i think

app/precompiled.py
```diff
@@ -234,9 +225,55 @@ def overlay_template_areas(src_pdf, page_number=None, overlay=True):
...
+            current_app.logger.warning('Letter size is not A4 on page {}, page size: {}x{}mm'.format(
+                page_num + 1, page_height, page_width
+            ))
+
+        # check if page orientation is not landscape:
+        if (
+            page_height < page_width and page.get('/Rotate') not in [90, 270]
+        ) or (page_height > page_width and page.get('/Rotate') not in [0, 180, None]):
+            invalid_pages.append(page_num + 1)
+            current_app.logger.warning("Letter landscape-oriented on page {}".format(page_num + 1))
```
leohemsted: is it worth logging height, width and /rotate here?

app/precompiled.py
```diff
@@ -234,9 +225,55 @@ def overlay_template_areas(src_pdf, page_number=None, overlay=True):
...
+            invalid_pages,
+            before_each='',
+            after_each='',
+            prefix='page',
+            prefix_plural='pages'
+        )
+    return (invalid_pages, message)
+
+
+def _are_pages_A4_portrait_oriented(src_pdf):
```
leohemsted: we can change this to `get_pages_with_invalid_orientation_or_size`, and then rather than returning a tuple, just return the list of invalid page numbers.

then we can just check if the list is empty to see whether it was okay or not

app/precompiled.py
CrystalPea: that looks shiny, I will apply it, thanks 👍

app/precompiled.py
CrystalPea: I don't think I understand... The current calculations give us size in milimeters, which is standard size of measuring A4, no?

app/precompiled.py
CrystalPea: more info, sounds good 👍

app/precompiled.py
CrystalPea: hmm true

app/precompiled.py
leohemsted: yeah, the calculation is correct (to go from points to mm), but rather than having magic numbers and doing the calculation ourselves, we've already got some shortcuts. elsewhere in precompiled.py we've got statements like `SOME_MEASUREMENT * mm` - `mm` is a hardcoded definition provided by `reportlab.units`, a module with a few hardcoded definitions for conversion

CrystalPea: ahhh, neat :3

CrystalPea: thanks, I've applied that now ☺️

app/precompiled.py
```diff
@@ -77,8 +77,8 @@ def sanitise_precompiled_letter():
         raise InvalidRequest('Sanitise failed - No encoded string')
 
     file_data = BytesIO(encoded_string)
-
-    if len(get_invalid_pages(file_data)) > 0:
+    invalid_pages, message = get_invalid_pages_with_message(file_data)
+    if len(invalid_pages) > 0:
         raise InvalidRequest('Sanitise failed - Document exceeds boundaries')
```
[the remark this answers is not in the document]
CrystalPea: you're right, thank you 💖

PR #357, opened 2019-08-01
move top of address block down

leohemsted:
the first line of the address is not our line, but from DVLA - it's a code they use to help route returned mail.

We're moving the top of the address block down from 39.5mm to 44mm to make sure that if someone tries to write content in that block, validation will fail.

TODO: Figure out strategy for deploying this - does it involve:

* Inform teams who use precompiled?
* Give them adequate time to update their letter templates - do we set a date in advance and publish that?


leohemsted:
This is no longer (as much of) an issue, because we now properly rewrite addresses, redact text and put a white box over it.

So any text that currently overlaps the box is no longer an issue.

PR #363, opened 2019-08-23
Redact pdf address block

CrystalPea:
Previously, we were using white rectangle to cover the address block
in precompiled letters sent by our users. Then we put properly formatted
address block on top of the white rectangle.
Unfortunately, the old address block was still machine-readable from
below the white rectangle, and it sometimes confused DVLA machines
that read postcodes and sort letters for sending.

Hence, we have to ensure that the address block is totally redacted
out of existence. This PR does that, with help of code from
[pdf_redactor](https://github.com/JoshData/pdf-redactor) github repository.

We now:
- find the address block in pdf's text stream and delete it.
- if we find more than one match, we do not redact, as this could
remove information form letter content block, and we log that we
found multiple matches for the address block.
- we also log if we did not find a match.
- we still use the white rectangle, in case one of the above
happens and hence we are unable to perform full redaction.

**BEFORE  |  AFTER**
Screen Shot 2019-08-23 at 15 16 18


yahoopete:
Will the existing code continue to re-write the address in the correct format after this happens? And side-question, does it CAPS everything up when it does that?

CrystalPea:
@yahoopete It all works fine, here is a screenshot of a file before and after redaction + rewriting:

Screen Shot 2019-08-23 at 16 36 12

They are almost identical only on 2nd one the address is a bit lower.

As to upper case, I don't think we change case of the characters in the process, unless I missed something @leohemsted ?

leohemsted:
@CrystalPea @yahoopete we don't currently UPPER CASE anything, but if we want to change that now it should be an easy fix

app/pdf_redactor.py
```diff
@@ -700,67 +588,79 @@ def update_text_layer(options, text_tokens, page_tokens):
...
-					# There will be more replaced later, so take the same number
-					# of characters from the replacement text.
-					r = replacement[:mlen]
-					replacement = replacement[mlen:]
-				else:
-					# This is the last token in which we'll replace text, so put
-					# all of the remaining replacement content here.
-					r = replacement
-					replacement = None # sanity
+		matches = [a for a in pattern.finditer(text_content)]
```
leohemsted: we don't need the iterator `find` method if we're just going to unwrap it. I think `matches = pattern.findall(text_context)` should work equivalent

app/precompiled.py
```diff
@@ -628,6 +631,23 @@ def is_notify_tag_present(pdf):
...
+        re.compile(address_regex),
+        lambda m: " "
+    ))
+    options.input_stream = pdf
+    options.output_stream = BytesIO()
+
+    pdf_redactor.redactor(options)
+
+    options.output_stream.seek(0)
+    return options.output_stream.read()
```
leohemsted: do you think we'll ever want greater control over what happens when we _don't_ redact text?

currently we log, but there's no way for this function to know if the address is still there for example.

requirements.txt
```diff
@@ -12,6 +12,8 @@ reportlab==3.5.6
 pdf2image==0.1.14
 PyMuPDF==1.13.20
 Werkzeug==0.14.1  # pyup: < 0.15.0
+pdfrw>=0.4
+defusedxml
```
leohemsted: we should pin these deps to versions, the same as above. just stick em both on the latest versions

app/pdf_redactor.py
CrystalPea: ahh, good point :)

app/precompiled.py
CrystalPea: >there's no way for this function to know if the address is still there for example

I'm not sure I understand. If we find no matches or multiple matches we are pretty confident that the function does not change anything right? So the old address would still be there, then covered by the white rectangle, then the formatted block would be put in front of it, no?

app/pdf_redactor.py
CrystalPea: Ahhh, I tried your suggestion and now I remember why I left the `finditer`. I tried `findall` before, too, but it returns a list of strings, whereas finditer gives  a colleciton of tuples which hold more information, like the position and length of the fragment of text.

leohemsted: aahhh 👍

app/precompiled.py
leohemsted: ah sorry, i meant more that the `redact_precompiled_letter_address_block` that calls the `pdf_redactor.redact` function doesn't know what happened.

For now that's fine because we're just accepting that they will have a doubled up (but covered) address block and will be unsorted. but if we want to change this or do anything else in the future, then it's going to be hard to do anything because we don't report anywhere whether the redaction process worked or not. All we'll have is a context-less log line.

It'd be nice if `pdf_redactor.redact` returned a boolean success flag or something, I think, even if we don't do anything with that for now.

CrystalPea: ahh, ok, thanks, that makes lots of sense, I will add it 👍

PR #367, opened 2019-09-04
Standarised response json for sanitise and validate endpoints

CrystalPea:
New standarised response json for sanitise and validate endpoints.

Those PRs need to go in first as they prepare our apps for the changes:
https://github.com/alphagov/notifications-api/pull/2604
https://github.com/alphagov/notifications-admin/pull/3118

Story ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/168057359

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -141,10 +141,9 @@ def init_app(app):
     @app.errorhandler(ValidationFailed)
     def validation_failed(error):
         return jsonify({
-            "validation_passed": False,
             "page_count": error.page_count,
             "recipient_address": None,
-            "errors": error.errors,
+            "message": error.message,
             "file": None
         }), error.code
 
@@ -182,7 +181,7 @@ class InvalidRequest(Exception):
 
 
 class ValidationFailed(Exception):
-    def __init__(self, errors, page_count=None, code=400):
-        self.errors = errors
+    def __init__(self, message, page_count=None, code=400):
+        self.message = message
         self.code = code
         self.page_count = page_count
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -133,11 +133,20 @@ def init_cache(application):
 
 
 def init_app(app):
-
     @app.errorhandler(InvalidRequest)
     def invalid_request(error):
         app.logger.warning(error.message)
         return jsonify(result='error', message=error.message or ""), error.code
+
+    @app.errorhandler(ValidationFailed)
+    def validation_failed(error):
+        return jsonify({
+            "validation_passed": False,
+            "page_count": error.page_count,
+            "recipient_address": None,
+            "errors": error.errors,
+            "file": None
+        }), error.code
 
     @app.errorhandler(Exception)
     def exception(error):
@@ -170,3 +179,10 @@ class InvalidRequest(Exception):
     def __init__(self, message, code=400):
         self.message = message
         self.code = code
+
+
+class ValidationFailed(Exception):
+    def __init__(self, errors, page_count=None, code=400):
+        self.errors = errors
+        self.code = code
+        self.page_count = page_count
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -9,7 +9,6 @@ from PIL import ImageFont
 from PyPDF2 import PdfFileWriter, PdfFileReader
 from flask import request, abort, send_file, Blueprint, jsonify, current_app
 from notifications_utils.statsd_decorators import statsd
-from notifications_utils.formatters import unescaped_formatted_list
 from pdf2image import convert_from_bytes
 from reportlab.lib.colors import white, black, Color
 from reportlab.lib.pagesizes import A4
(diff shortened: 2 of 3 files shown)


app/precompiled.py
```diff
@@ -229,26 +237,24 @@ def overlay_template_areas(src_pdf, page_number=None, overlay=True):
...
-            before_each='',
-            after_each='',
-            prefix='page',
-            prefix_plural='pages'
-        )
-    return (invalid_pages, message)
+        errors["content_outside_of_printable_area"] = list(_get_out_of_bounds_pages(PdfFileReader(pdf_to_validate)))
+        if not errors["content_outside_of_printable_area"]:
+            passed_validation = True
+    return passed_validation, errors
```
leohemsted: dont think you really need to keep track of this `passed_validation` variable - you can just check if any of the lists have data in when returning.

 i think this'll work: `return not any(errors.values()), errors`. though that looks a bit magic, unsure how it'll look.

tests/test_precompiled.py
```diff
@@ -762,10 +790,10 @@ def test_precompiled_sanitise_pdf_without_notify_tag(client, auth_header):
             **auth_header
         }
     )
-
     assert response.status_code == 200
+    json_data = json.loads(response.get_data())
```
leohemsted: can we add checks to this test `test_precompiled_sanitise_pdf_without_notify_tag` and `test_precompiled_validation_endpoint_blank_pdf` (as the two simplest happy path tests i think) that just assert the entire json layout? just so we can be sure that the errors dict is empty, page count is populated, etc?

app/precompiled.py
```diff
@@ -240,14 +240,14 @@ def overlay_template_areas(src_pdf, page_number=None, overlay=True):
 def get_invalid_pages_with_errors(src_pdf):
     passed_validation = False
     errors = {
-        "pages_out_of_printable_area": [],
-        "non_a4_portrait_pages": [],
+        "content_outside_of_printable_area": [],
+        "document_not_a4_size_portrait_orientation": [],
```
leohemsted: in fact, after looking through i prefer the old values as keys - makes it clear it's talking about pages. (appreciate you changed this for admin's sake, but i'd rather admin just had a mapping dict.

app/precompiled.py
```diff
@@ -80,19 +79,28 @@ def sanitise_precompiled_letter():
...
     if not does_pdf_contain_cmyk(encoded_string) or does_pdf_contain_rgb(encoded_string):
         file_data = BytesIO(convert_pdf_to_cmyk(encoded_string))
     if not is_notify_tag_present(file_data):
         file_data = add_notify_tag_to_letter(file_data)
 
-    file_data = rewrite_address_block(file_data)
+    file_data, recipient_address = rewrite_address_block(file_data)
 
-    return send_file(filename_or_fp=file_data, mimetype='application/pdf')
+    # return send_file(filename_or_fp=file_data, mimetype='application/pdf')
```
leohemsted: can remove this commented out line

app/__init__.py
```diff
@@ -170,3 +179,10 @@ class InvalidRequest(Exception):
     def __init__(self, message, code=400):
         self.message = message
         self.code = code
+
+
+class ValidationFailed(Exception):
+    def __init__(self, errors, page_count=None, code=400):
+        self.errors = errors
```
leohemsted: just a thought, it might be good to define the dict containing all the possible errors in one place. That way, if we add new functions that do different validation checks, there's still one place that contains the list of all values, and each function doesn't need to worry about all the other values that other functions might add. This json is a boundary contract so good to be explicit about it in an easily located place 🙂

```python
self.errors = {
    "content_outside_of_printable_area": [],
    "document_not_a4_size_portrait_orientation": [],
}

for key, value in errors:
    if key in self.errors:
        self.errors[key] = value
```

leohemsted:
also, what does the response look like when the file is invalid, or can't be read for some reason?

servingUpAces:
This feels a bit complex. We have a boolean to say if the validation passed and errors - which is populated whether or not we have errors. Then the admin app needs to do something with the error message. I'm happy to chat about this since I haven't gone as far as figure out how I would make it less complex.

CrystalPea:
> also, what does the response look like when the file is invalid, or can't be read for some reason?

@leohemsted I think that bit did not change so far, so it would be:
`return jsonify(result='error', message=error.message or ""), error.code`

Also @klssmith put my attention to those cases where file is an encoded string but is missing an EOF marker - currently we do not handle those and they cause an error I think. Should we start handling them? And if yes, should it be part of this PR or maybe better a separate one? (I feel it would be better to do it separately to not overcomplicate things)

tests/test_precompiled.py
CrystalPea: I added the check to `test_precompiled_sanitise_pdf_without_notify_tag`, but not to the other test since the other test uses validate endpoint that will become obsolete when we ship single file upload through the interface, as we talked on Thursday. I hope this is ok 👍

app/precompiled.py
```diff
@@ -119,21 +125,21 @@ def validate_pdf_document():
     if not encoded_string:
         abort(400)
 
-    invalid_pages, message = get_invalid_pages_with_message(BytesIO(encoded_string))
+    message = get_invalid_pages_with_message(BytesIO(encoded_string))
     data = {
-        'result': len(invalid_pages) == 0,
+        'result': bool(not message)
```
servingUpAces: Does it make sense that result is True or False instead of Success or Fail?

CrystalPea: Success or Fail might be better, but I think ultimately it's not worth changing it now since we want to drop that endpoint soon anyway (it's the one only used by Platform Admins currently, and when we deploy single upload, it will be obsolete, right?)

PR #369, opened 2019-09-09
Let API know when full redaction fails, so we can check source file

CrystalPea:


app/pdf_redactor.py
```diff
@@ -107,6 +107,8 @@ def redactor(options):
 	writer = PdfWriter()
 	writer.trailer = document
 	writer.write(options.output_stream)
+	if message:
+		return message
```
leohemsted: you should assign message, or if `options.content_filters` isn't supplied it'll error because message isn't defined.

If you just set it to None then we can always `return message`

app/pdf_redactor.py
```diff
@@ -601,16 +603,14 @@ def update_text_layer(options, text_tokens, page_tokens):
...
+			return message
 
 		if len(matches) > 1:
-			current_app.logger.warning(
-                "More than one match for address block during redaction procedure"
-            )
-			return
+			message = "More than one match for address block during redaction procedure"
+			current_app.logger.warning(message)
+			return message
```
leohemsted: we should add `return None` to the end of this function (line 667) just to be explicit

PR #370, opened 2019-09-10
Squash unexpected pdf exceptions

leohemsted:
move ValidationFailed error handler to generic precompiled one. If validation fails unexpectedly, we should handle this the same as if it cannot open the PDF.

the /precompiled/ blueprint currently has four endpoints:

* /precompiled/sanitise
* /precompiled/add_tag
* /precompiled/validate
* /precompiled/overlay.<file_type>

add_tag isn't used and validate is going away soon (when the one-off pdf epic is shipped), so we don't need to worry about them. overlay is used to display images that have already been through the validation process so we know they are at least processable.

So really, sanitise is the only endpoint that looks at PDFs for the first time, running the risk of being unable to open them. We've seen errors with PyPDF2 where it raised a NoneTypeError rather than a regular validation error, so we're now catching all `Exception` errors in a blueprint level handler (so it doesn't interfere with exceptions in other endpoints).

It looks for `page_count`, `message` and `code` attributes on exceptions, but otherwise returns the same json as the existing ValidationFailed error handler.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,3 +8,4 @@ pytest-cov==2.7.1
 pytest-mock==1.10.4
 pytest==4.6.2
 requests-mock==1.6.0
+pdbpp
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,4 +8,4 @@ pytest-cov==2.7.1
 pytest-mock==1.10.4
 pytest==4.6.2
 requests-mock==1.6.0
-pdbpp
+pdbpp==0.10.0
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,7 +15,6 @@ def app():
 
 @pytest.fixture
 def client(app):
-    app.config['DEBUG'] = True
     app.config['TESTING'] = True
 
     with app.test_request_context(), app.test_client() as client:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,9 @@ def app():
 
 @pytest.fixture
 def client(app):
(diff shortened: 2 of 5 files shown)


tests/conftest.py
```diff
@@ -15,6 +15,9 @@ def app():
 
 @pytest.fixture
 def client(app):
+    app.config['DEBUG'] = True
+    app.config['TESTING'] = True
```
CrystalPea: what do those variables do?

requirements-dev.txt
```diff
@@ -8,3 +8,4 @@ pytest-cov==2.7.1
 pytest-mock==1.10.4
 pytest==4.6.2
 requests-mock==1.6.0
+pdbpp
```
CrystalPea: was pdbpp added just temporarily or do we want it for good? If we want it for good, don't we need to pin version?

tests/conftest.py
leohemsted: https://flask.palletsprojects.com/en/1.1.x/config/#DEBUG - probably doesn't actually do anything in this context, and I reckon we can remove this
https://flask.palletsprojects.com/en/1.1.x/config/#TESTING - disables error handler (so if theres an exception you see the stack trace rather than just see the function returning a 500 error page)

requirements-dev.txt
leohemsted: i'll pin version. it's a useful improvement over pdb

tests/conftest.py
CrystalPea: cool, thanks 👍

PR #371, opened 2019-09-11
bump requirements (fix logo cropping)

leohemsted:
notable bumps:

* docker image from stretch to buster
* weasyprint from 40 to 51
* werkzeug from 0.14 to 1.0
* pillow from 5.x to 7.x
* pdf2image from 0.1 to 1.12.x
* remove `libcairo2` as an explicit dependency. This means it ends up being bumped from 1.14  1.16.0-4. 

notable pins:

* cairosvg 2.4.x, when converting svgs to pngs, makes them tiny (as it no longer accepts _just_ a width arg, it needs both width and height. Solved this by not using the PNGs at all.


requirements.txt
```diff
@@ -1,28 +1,29 @@
...
+git+https://github.com/alphagov/notifications-utils.git@36.3.1#egg=notifications-utils==36.3.1
 
 # PaaS requirements
-gunicorn==19.9.0
+gunicorn==20.0.4
 awscli-cwlogs>=1.4,<1.5
+
+# CairoSVG-2.4.2
+# Flask-1.1.1
+# Werkzeug-0.16.0
```
quis: Why are these commented out?

Flask and Werkzeug seem to be duplicates of definitions above. The commented out version of CairoSVG is newer than the one referenced above.

leohemsted: it's been a while since i looked at this but iirc these versions were crashing. i might take a look at this PR again later today/tomorrow. either way these commented lines shouldn't really be here.

requirements.txt
```diff
@@ -1,28 +1,26 @@
-# app requirements
+#app requirements
+celery[sqs]==4.4.1
+Flask==1.1.1
+Flask-WeasyPrint==0.6
+Flask-HTTPAuth==3.3.0
+Werkzeug==1.0.0
+
+# pdf libraries
 CairoSVG==2.2.1
```
leohemsted: version `2.4.2` is out now, but that version breaks svg -> png conversion

we no longer use that code path anyway - the PNG version of the logo is only used in templated letter print previews, which we have changed to now use the vector version.

CrystalPea:
Hey,

I've been checking the flows:
- templated one-off flow looks good for a correct letter, both png and pdf look good
- uploading a correct precompiled letter via both API and uploads works well, and previews do look well, too.

But there is a thing that this PR broke, and that is validation messages for precompiled letters, both via API and uploads:

### Uploads
Screenshot 2020-03-03 at 11 16 06
Screenshot 2020-03-03 at 11 15 36

### API
Screenshot 2020-03-03 at 11 24 13
Screenshot 2020-03-03 at 11 24 02

PR #373, opened 2019-09-12
Steps to make precompiled redaction work better

CrystalPea:
- move redaction above adding notify tag and conversion to CMYK, because pdfrw struggled to read the pdf made by us - it only read Notify tag layer for some pdfs
-  change text extraction function so it uses fitz library. Previous function had a problem with recognising some weirdly formatted whitespace characters which affected both redaction and the rewritten address.

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -109,10 +109,11 @@ def sanitise_precompiled_letter():
     if message:
         raise ValidationFailed(message, page_count=page_count)
 
-    file_data, recipient_address, redaction_failed_message = rewrite_address_block(file_data)
     # during switchover, DWP will still be sending the notify tag. Only add it if it's not already there
     if not does_pdf_contain_cmyk(encoded_string) or does_pdf_contain_rgb(encoded_string):
         file_data = BytesIO(convert_pdf_to_cmyk(encoded_string))
+
+    file_data, recipient_address, redaction_failed_message = rewrite_address_block(file_data)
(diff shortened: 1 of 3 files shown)


app/precompiled.py
```diff
@@ -582,70 +586,54 @@ def rewrite_address_block(pdf):
...
+    doc = fitz.open("pdf", pdf)
+    page = doc[0]
+    rect = fitz.Rect(x1, y1, x2, y2)
+    words = page.getTextWords()
+    mywords = [w for w in words if fitz.Rect(w[:4]).intersects(rect)]
+    mywords.sort(key=itemgetter(3, 0))
+    group = groupby(mywords, key=itemgetter(3))
+    address = []
+    for y1, gwords in group:
+        address.append(" ".join(w[4] for w in gwords))
```
leohemsted: what's going on here? Did you take this snippet from docs/stackoverflow or anything?

Can you put in some comments or links to relevant docs/examples etc to explain what's happening with all these itemgetters and array accesses?

CrystalPea: That's a fair point, I took the script from here:
https://github.com/pymupdf/PyMuPDF-Utilities/blob/master/textboxtract.py

And it was referenced in the library docs here: https://pymupdf.readthedocs.io/en/latest/faq/#how-to-extract-text-from-within-a-rectangle

Should I add this information to the function's docstring?

PR #395, opened 2019-11-29
Tidy up after move to concourse

idavidmcdonald:
Review commit by commit

Note, this is not perfect. The steps to manually deploy this don't set the docker image tag as the short git SHA like we do on concourse.

Worth taking a look at our current status page:
https://preview-notify-template-preview.cloudapps.digital/_status

It will now show the correct git commit and drop non useful keys.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ TEMPLATE_PREVIEW_API_KEY ?= "my-secret-key"
 DOCKER_CONTAINER_PREFIX = ${USER}-${BUILD_TAG}
 
 NOTIFY_CREDENTIALS ?= ~/.notify-credentials
-CF_MANIFEST_FILE ?= manifest-${CF_SPACE}.yml
 
 NOTIFY_APP_NAME ?= notify-template-preview
 CF_APP = notify-template-preview
--- a/Makefile
+++ b/Makefile
@@ -4,16 +4,11 @@ DATE = $(shell date +%Y-%m-%dT%H:%M:%S)
 
 APP_VERSION_FILE = app/version.py
 
-GIT_COMMIT ?= $(shell git rev-parse HEAD 2> /dev/null || cat commit || echo "")
-
-BUILD_TAG ?= notifications-template-preview-manual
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -58,45 +55,9 @@ curl \
 
 ## Deploying
 
-You shouldn’t need to deploy this manually because there’s a pipeline setup in Jenkins. If you do want to deploy it manually, you'll need the notify-credentials repo set up locally.
+You shouldn’t need to deploy this manually because there’s a pipeline setup in Concourse. If you do want to deploy it manually, you'll need the notify-credentials repo set up locally.
 
 ```shell
 make (sandbox|preview|staging|production) upload-to-dockerhub
 make (sandbox|preview|staging|production) cf-deploy
```
klssmith: I removed the sandbox target from the Makefile recently but forgot to remove references to it from the README, so could take out anything that mentions the sandbox environment from here since we're tidying it up

Makefile
```diff
@@ -16,24 +16,18 @@ TEMPLATE_PREVIEW_API_KEY ?= "my-secret-key"
...
 CF_ORG ?= govuk-notify
-CF_HOME ?= ${HOME}
-$(eval export CF_HOME)
 CF_SPACE ?= development
 
 DOCKER_IMAGE = govuknotify/notifications-template-preview
 DOCKER_IMAGE_TAG = ${DEPLOY_BUILD_NUMBER}
 
 DOCKER_IMAGE_NAME = ${DOCKER_IMAGE}:${DOCKER_IMAGE_TAG}
-DOCKER_TTY ?= $(if ${JENKINS_HOME},,t)
```
klssmith: Removing this has the effect of not attaching a tty to the container when running locally, which means you can't stop the the app running in Docker with ctrl-c anymore. Not sure of there's a simple way of getting that behaviour back, because it's quite handy being able to stop the container with ctrl-c locally

idavidmcdonald: good spot! thanks

README.md
idavidmcdonald: Yes, I can do this.

PR #399, opened 2019-12-10
Add a Celery app for sanitising letters

klssmith:
We currently create a synchronous https request to template-preview when we need to validate letters but this can create large numbers of requests on the app, causing it to become overloaded and fail. This PR adds a new Celery app for template-preview so that we can put the validation of letters on a template-preview queue.

See commit messages for details, but this essentially:
- Adds the Celery config and new task
- Refactors the `/precompiled/sanitise` endpoint so that it can be used in an HTTP request and task
- Makes some changes to the Dockerfile and Makefile so that they work with both apps

[Pivotal story](https://www.pivotaltracker.com/story/show/169148522)

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -37,7 +37,7 @@ def load_config(application):
         'timezone': 'Europe/London',
         'imports': ['app.celery.tasks'],
         'task_queues': [
-            Queue(QueueNames.TEMPLATE_PREVIEW, Exchange('default'), routing_key=QueueNames.TEMPLATE_PREVIEW)
+            Queue(QueueNames.SANITISE_LETTERS, Exchange('default'), routing_key=QueueNames.SANITISE_LETTERS)
         ],
     }
 
@@ -218,7 +218,7 @@ class ValidationFailed(Exception):
 
 class QueueNames:
     LETTERS = 'letter-tasks'
-    TEMPLATE_PREVIEW = 'template-preview-tasks'
+    SANITISE_LETTERS = 'sanitise-letter-tasks'
 
 
 class TaskNames:
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -63,7 +63,7 @@ def load_config(application):
             application.config['NOTIFY_ENVIRONMENT']
         )
     )
-    application.config['S3_LETTER_CACHE_BUCKET'] = (
+    application.config['LETTER_CACHE_BUCKET_NAME'] = (
         '{}-template-preview-cache'.format(
             application.config['NOTIFY_ENVIRONMENT']
         )
@@ -142,7 +142,7 @@ def init_cache(application):
 
                 with suppress(S3ObjectNotFound):
                     return s3download(
-                        application.config['S3_LETTER_CACHE_BUCKET'],
+                        application.config['LETTER_CACHE_BUCKET_NAME'],
                         cache_key,
                     )
 
@@ -151,7 +151,7 @@ def init_cache(application):
                 s3upload(
                     data,
                     application.config['AWS_REGION'],
-                    application.config['S3_LETTER_CACHE_BUCKET'],
+                    application.config['LETTER_CACHE_BUCKET_NAME'],
                     cache_key,
                 )
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -5,6 +5,7 @@ from hashlib import sha1
 
 import PyPDF2
 import binascii
+from kombu import Exchange, Queue
 from weasyprint.logger import LOGGER as weasyprint_logs
 from flask import Flask, jsonify, abort
 from flask_httpauth import HTTPTokenAuth
@@ -13,11 +14,33 @@ from notifications_utils import logging as utils_logging
 from notifications_utils.clients.statsd.statsd_client import StatsdClient
 from notifications_utils.s3 import s3upload, s3download, S3ObjectNotFound
 
+from app.celery.celery import NotifyCelery
+
+
+notify_celery = NotifyCelery()
+
 
 def load_config(application):
+    application.config['AWS_REGION'] = 'eu-west-1'
     application.config['API_KEY'] = os.environ['TEMPLATE_PREVIEW_API_KEY']
     application.config['NOTIFY_ENVIRONMENT'] = os.environ['NOTIFY_ENVIRONMENT']
     application.config['NOTIFY_APP_NAME'] = 'template-preview'
+
+    application.config['BROKER_URL'] = 'sqs://'
+    application.config['BROKER_TRANSPORT_OPTIONS'] = {
+        'region': application.config['AWS_REGION'],
+        'polling_interval': 1,
+        'visibility_timeout': 310,
+        'queue_name_prefix': get_queue_prefix(application.config['NOTIFY_ENVIRONMENT']),
+    }
+    application.config['CELERY_ENABLE_UTC'] = True
+    application.config['CELERY_TIMEZONE'] = 'Europe/London'
+    application.config['CELERY_ACCEPT_CONTENT'] = ['json']
+    application.config['CELERY_TASK_SERIALIZER'] = 'json'
+    application.config['CELERY_IMPORTS'] = ['app.celery.tasks']
+    application.config['CELERY_QUEUES'] = [
+        Queue(QueueNames.TEMPLATE_PREVIEW, Exchange('default'), routing_key=QueueNames.TEMPLATE_PREVIEW)
+    ]
 
     # if we use .get() for cases that it is not setup
     # it will still create the config key with None value causing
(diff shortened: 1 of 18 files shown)


app/celery/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@notify_celery.task(name='sanitise-and-upload-letter')
+@statsd(namespace='template-preview')
+def sanitise_and_upload_letter(notification_id, filename):
+    current_app.logger.info('Sanitising notification with id {}'.format(notification_id))
+
+    try:
+        pdf_content = s3download(current_app.config['LETTERS_SCAN_BUCKET_NAME'], filename).read()
+        json_response = sanitise_file_contents(pdf_content)
```
idavidmcdonald: Discussed whether json_response sounds too much like to do with an http request, could maybe call something like sanitisation_details

app/celery/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+
+        # Only files that have failed validation have a message in the json_response
```
idavidmcdonald: Discussed maybe changing comment to be consistent around `sanitisation` rather than `validation`

app/celery/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+
+            # If the file already exists in S3, it will be overwritten
+            s3upload(
+                filedata=file_data,
+                region=current_app.config['AWS_REGION'],
+                bucket_name=current_app.config['S3_LETTER_SANITISE_BUCKET'],
+                file_location=filename,
+            )
+
+        current_app.logger.info('Notification {} validation: {}'.format(validation_status, notification_id))
```
idavidmcdonald: Mention the word sanitisation in this log line

app/celery/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+            validation_status = 'failed'
+        else:
+            validation_status = 'passed'
+            file_data = base64.b64decode(json_response['file'].encode())
+
+            # If the file already exists in S3, it will be overwritten
+            s3upload(
+                filedata=file_data,
+                region=current_app.config['AWS_REGION'],
+                bucket_name=current_app.config['S3_LETTER_SANITISE_BUCKET'],
```
idavidmcdonald: This sounds like a bucket of things to be sanitised rather than something that has been sanitised

tests/test_precompiled.py
```diff
@@ -200,7 +200,7 @@ def test_precompiled_validation_with_preview_throws_error_if_file_is_not_a_pdf(c
 
     assert response.status_code == 400
     json_data = json.loads(response.get_data())
-    assert json_data['message'] == 'unable-to-read-the-file'
+    assert json_data['message'] == 'Unable to read the PDF data: Could not read malformed PDF file'
```
idavidmcdonald: I couldn't quite work out why this has changed. I'm vaguely on the right lines but couldn't identify the exact line of code that changes the result. If you could point it out to me in person tomorrow that would be great, thanks!

manifest-celery.yml.j2
```diff
@@ -0,0 +1,29 @@
...
+  disk_quota: 2G
+  memory: 2G
+  health-check-type: process
+  command: exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=5 --uid=`id -u celeryuser`
+
+  services:
+    - logit-ssl-syslog-drain
+
+  routes:
+    - route: {{ environment }}-notify-template-preview-celery.cloudapps.digital
```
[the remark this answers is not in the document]
klssmith: I think the reason that template-preview is the only app to have the environment at the beginning of the route is to avoid routes called `notify-template-preview-preview` or `notify-template-preview-production` which might be a bit unclear.

Keeping the route as `preview-notify-template-preview-celery` seems maybe(?) a bit clearer than `notify-template-preview-celery-preview`.

app/celery/tasks.py
klssmith: Have updated this name and the other places in the task where it would make things clearer to use the word / name `sanitise`

app/__init__.py
```diff
@@ -35,13 +58,23 @@ def load_config(application):
     else:
         application.config['STATSD_ENABLED'] = False
 
-    application.config['S3_REGION'] = 'eu-west-1'
+    application.config['LETTERS_SCAN_BUCKET_NAME'] = (
```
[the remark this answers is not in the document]
klssmith: We use `LETTERS_SCAN_BUCKET_NAME` in both notifications-api and notifications-antivirus, so I've kept it the same here but have changed the other two names instead (to `LETTER_CACHE_BUCKET_NAME` and `SANITISED_LETTER_BUCKET_NAME`).

app/celery/tasks.py
klssmith: 👍Updated to `SANITISED_LETTER_BUCKET_NAME`

docker/Dockerfile
```diff
@@ -66,23 +66,21 @@ COPY . .
...
 RUN \
 	echo "Installing python dependencies" \
 	&& pip install -r requirements.txt
 
 COPY app app
 COPY wsgi.py gunicorn_config.py Makefile ./
-COPY scripts/run_app_paas.sh scripts/run_app.sh scripts/
-
+COPY scripts/run_app_paas.sh scripts/run_app.sh scripts/run_celery.sh scripts/
+COPY run_celery.py .
```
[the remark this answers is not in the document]
klssmith: `run_celery.py` is not in `scripts/` (only `run_celery.sh` is) so I'm not sure how to add it to line 80

app/__init__.py
```diff
@@ -174,3 +209,20 @@ def __init__(self, message, invalid_pages=None, page_count=None, code=400):
...
+    LETTERS = 'letter-tasks'
+    TEMPLATE_PREVIEW = 'template-preview-tasks'
+
+
+class TaskNames:
+    PROCESS_SANITISED_LETTER = 'process-sanitised-letter'
+
+
+queue_prefix = {
+    'development': os.environ.get('NOTIFICATION_QUEUE_PREFIX', 'development'),
```
leohemsted: it'd be nice if it throw an exception if your environment is `development` and your NOTIFICATION_QUEUE_PREFIX is unset. Might need this to be made in to a `get_queue_prefix(environment)` function though

manifest-celery.yml.j2
```diff
@@ -0,0 +1,29 @@
...
+
+
+  env:
+    NOTIFY_ENVIRONMENT: {{ environment }}
+
+    NOTIFY_APP_NAME: template-preview-celery
```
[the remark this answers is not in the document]
klssmith: Good question - I had a look at this and notifications-api has different values of `CW_APP_NAME` for each of its apps, so we do different things in different repos. Going to keep this as different here so that we get different log groups per app.

app/celery/celery.py
```diff
@@ -0,0 +1,30 @@
...
+        # this task is nested so that it has access to the original flask application - when celery restarts processes,
+        # it doesn't appear to execute the whole import process again, and as such, flask.current_app no longer has
+        # a context, and throws a RuntimeError. So we need to create an app context from scratch each time.
+        class NotifyTask(Task):
+            abstract = True
+
+            def on_failure(self, exc, task_id, args, kwargs, einfo):
+                # ensure task will log exceptions to correct handlers
+                with app.app_context():
+                    app.logger.exception('Celery task failed')
```
leohemsted: can you chuck `self.name` in to this log message to make debugging a lil' quicker?

eg `Celery task {self.name} failed`

requirements.txt
```diff
@@ -1,5 +1,6 @@
 # app requirements
 CairoSVG==2.2.1
+celery==3.1.26.post2 # pyup: <4
```
leohemsted: celery 3 😢

can we just quietly put this on celery 4? does that work? (i know that 3.1.26 and 4 were cross-compatible when i tried)

app/celery/celery.py
klssmith: Good suggestion, I've made that change

app/__init__.py
klssmith: I've made a function to check (rebased into the first commit). The downside is that you now need to have `NOTIFICATION_QUEUE_PREFIX` set for both apps, even though the Flask app doesn't use it. But I guess that's probably ok

requirements.txt
klssmith: Changed to Celery 4 in d716f29. It worked fine locally, but I guess we may want to do a load test on staging when this is deployed.

app/__init__.py
klssmith: On second thoughts, having tried this a bit more I'm not sure that it is an improvement to throw an exception when `NOTIFICATION_QUEUE_PREFIX` is unset. The problem is that when you run the Flask app it raises the error, but the message about queue prefix not being set is pretty far up the stack trace and at first glance it looks like the issue is a Gunicorn one. So it's quite hard to tell what the problem is when you're running the Flask app

Might be better to undo this change and just default to a queue prefix of `development` like we do in api and antivirus

klssmith: Changed here https://github.com/alphagov/notifications-template-preview/pull/399/commits/9bc308ee1adf4cf2c7cf76dfe3b9bd1ebcd1a316 to the approach we talked about of having the Makefile check that the environment variable is set. `make run-with-docker` (to run the Flask app) doesn't give any errors if `NOTIFICATION_QUEUE_PREFIX` is unset, so we don't need to add a default value.

docker/Dockerfile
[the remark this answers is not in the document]
idavidmcdonald: So line 80 would look like: `COPY wsgi.py gunicorn_config.py Makefile run_celery.py ./`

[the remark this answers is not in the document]
klssmith: That makes sense, I've updated this now (read the line numbers wrong initially and thought you were suggesting adding it to the line starting `COPY scripts/run_app_paas.sh ...`

PR #407, opened 2020-01-20
Make margins more generous

leohemsted:
when we validate, we use margins provided by DVLA to validate against to prevent people writing in places where either

* machines can't print on the edges of the paper
* DVLA will put barcodes and things that we shouldn't cover up

However we don't need to be as precise on these borders. Utimately, a few pixels just over a border will be fine. This might happen because someone's aligned text, but the text goes outside the normal bounding box (eg the top left of a letter W). Also, when we check boundaries we convert to an image and then check the pixel colours - this may cause issues if anti-aliasing puts a fuzzy grey pixel just over.

Add a 1mm margin-of-error to each border so that if someone puts a letter right on the edge, it won't trip our validation. This change also means that if the validation does fail, the red boundaries they see on their failed pdf will also be pushed out 1mm. However, we're deliberately not updating the guidance template so that they still target the original DVLA margins.

One question I've not figured out the answer to is do we want these margins to be shown as their more generous variants when they get a red overlay? We validate against softer boundaries but I wonder if we show them those softer boundaries they might try and align against those and the problem won't be solved?


yahoopete:
Strong reckon that we continue to show the original overlay so that it'll be easier to spot when there is an incursion. If we keep moving the overlay, think we'll just move the problem to the new edge....

app/precompiled.py
```diff
@@ -37,27 +37,39 @@
 FONT = "Arial"
 TRUE_TYPE_FONT_FILE = FONT + ".ttf"
 
-BORDER_FROM_BOTTOM_OF_PAGE = 5.0
-BORDER_FROM_TOP_OF_PAGE = 5.0
-BORDER_FROM_LEFT_OF_PAGE = 15.0
-BORDER_FROM_RIGHT_OF_PAGE = 15.0
-BODY_TOP_FROM_TOP_OF_PAGE = 95.00
+BORDER_FROM_BOTTOM_OF_PAGE = 5.0 - 1.0
```
idavidmcdonald: I think a code comment, probably here, to say what the +/- 1.0s are for generally in all the following variables. But I think I like the +/- 1.0 approach over say just changing the number to 4.0.

leohemsted:
Closing this - this approach didn't lend itself well to having different values for validation vs showing errors to users.

PR #408, opened 2020-01-24
Remove old deprecated functions

leohemsted:
everything looks the same as it used to locally

image
image

tried to split this out into sensible commits, succeeded to some degree with that.

* separate overlaying with white for validation and colouring with red for showing to users
* removed the old validate endpoint that isn't used
* moved relevant validate tests that used to hit the old endpoint to just call the relevant methods
* tweaked some test PDFs to make them a bit more useful. there's lots more work that can be done here.
* added a (marginally) more realistic valid pdf.
* tried to give IDs to tests that take in PDFs, so if they fail pytest doesn't print out the binary data

***

Closes #396


app/precompiled.py
```diff
@@ -138,80 +138,35 @@ def sanitise_file_contents(encoded_string):
...
         abort(400)
 
     file_data = BytesIO(encoded_string)
 
     if file_type == 'png':
+        page = request.args.get('page_number')
+        if not page:
+            current_app.logger.error(f'expected page_number in {request.args}')
+            # this endpoint can only be called with a zero-indexed page number
+            abort(400)
```
servingUpAces: using `abort(400)` isn't very useful for the calling app. At the moment we aren't handling 400s, but maybe we should at least show a decent message. (this is just a comment - I realise this PR is not about that and don't expect any changes now)

app/precompiled.py
```diff
@@ -523,6 +425,34 @@ def _colour_no_print_areas(src_pdf, page_number=1):
         can.rect(page_width - right, 0, page_width, page_height, fill=True, stroke=False)
         # bottom margin:
         can.rect(left, 0, page_width - (2 * right), bottom, fill=True, stroke=False)
+
+        # The first page is more varied because of address blocks etc subsequent pages are more simple
+        if page_num == 0:
```
servingUpAces: In the deleted code `if page_num == 1` now it's 0, why is that?

leohemsted: the deleted code was looking at a user provided page_num variable, which starts at page 1

this is defined in `for page_num in range(0, pdf.numPages)`, and so is 0 indexed

PR #409, opened 2020-01-24
Add letter address to celery task data

klssmith:
The `sanitise_and_upload_letter` Celery task now calls the `process_sanitised_letter` task in notifications-api with a single argument of encrypted data, instead of with kwargs of unencrypted data. The letter address is also added to the data sent to api.

To encrypt the data, this brings in the encryption client from the latest version of utils and adds two environment variables - `DANGEROUS_SALT` and `SECRET_KEY`.

Since the `process_sanitised_letter` task in api is not currently used anywhere, this change doesn't need to be backwards compatible.

See https://github.com/alphagov/notifications-api/pull/2699

[Pivotal story](https://www.pivotaltracker.com/story/show/170649957)

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -25,6 +25,8 @@ def load_config(application):
     application.config['API_KEY'] = os.environ['TEMPLATE_PREVIEW_API_KEY']
     application.config['NOTIFY_ENVIRONMENT'] = os.environ['NOTIFY_ENVIRONMENT']
     application.config['NOTIFY_APP_NAME'] = 'template-preview'
+    application.config['DANGEROUS_SALT'] = os.environ['DANGEROUS_SALT']
+    application.config['SECRET_KEY'] = os.environ['SECRET_KEY']
 
     application.config['celery'] = {
         'broker_url': 'sqs://',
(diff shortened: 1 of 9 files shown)


app/__init__.py
```diff
@@ -25,6 +26,8 @@ def load_config(application):
     application.config['API_KEY'] = os.environ['TEMPLATE_PREVIEW_API_KEY']
     application.config['NOTIFY_ENVIRONMENT'] = os.environ['NOTIFY_ENVIRONMENT']
     application.config['NOTIFY_APP_NAME'] = 'template-preview'
+    application.config['DANGEROUS_SALT'] = os.environ['DANGEROUS_SALT']
+    application.config['SECRET_KEY'] = os.environ['SECRET_KEY']
```
klssmith: We don't need these new environment variables for the Flask app, only the Celery app, so this could be changed to `os.environ.get('SECRET_KEY')` instead so that we can leave the variables out of the Flask manifest. Doing things this way seemed slightly safer since it ensures that the variables must always be set for the Celery app.

servingUpAces: I think I'm okay with leaving this as is. If someone decides to use the encryption class, at least they are using the right keys.

PR #413, opened 2020-01-29
Refactor canvas functions

leohemsted:
Best reviewed commit-by-commit.

The goal of this was to:

* reduce boilerplate
* standardise our static variables.
* reduce the amount of maths that takes place in the drawing functions

PR #414, opened 2020-01-29
Move margins

leohemsted:
- [x] https://github.com/alphagov/notifications-template-preview/pull/413 (split out to make it easier to review)
- [ ] this pr

on prod it is rejected: 
image

on my local it works:
image

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -337,23 +337,23 @@ def _overlay_printable_areas_with_white(src_pdf):
     # The first page is more varied because of address blocks etc subsequent pages are more simple
 
     # Body
-    pt1 = BORDER_LEFT_FROM_LEFT_OF_PAGE, BODY_TOP_FROM_TOP_OF_PAGE
-    pt2 = BORDER_RIGHT_FROM_LEFT_OF_PAGE, BORDER_BOTTOM_FROM_TOP_OF_PAGE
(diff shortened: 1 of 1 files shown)


app/precompiled.py
```diff
@@ -337,23 +337,23 @@ def _overlay_printable_areas_with_white(src_pdf):
     # The first page is more varied because of address blocks etc subsequent pages are more simple
 
     # Body
-    pt1 = BORDER_LEFT_FROM_LEFT_OF_PAGE, BODY_TOP_FROM_TOP_OF_PAGE
-    pt2 = BORDER_RIGHT_FROM_LEFT_OF_PAGE, BORDER_BOTTOM_FROM_TOP_OF_PAGE
+    pt1 = BORDER_LEFT_FROM_LEFT_OF_PAGE - 1, BODY_TOP_FROM_TOP_OF_PAGE - 1
```
CrystalPea: is there a reason we are changing those values in the functions, not in the constants at the top?

leohemsted: the constants at the top are used for both

* validating that content is in bounds
* displaying the red rectangle that people see when their validation fails.

if we change the constants at the top, then the boundaries will move out for both of these functions. We reckon that people will line up their content with the red borders, so they should stay tight or we won't have achieved anything.

CrystalPea: Then why are we adding and subtracting those points at all? If we do it, won't our validation be slightly different from our overlay? Why would be want that? I'm confused 😕

PR #418, opened 2020-02-20
Embed fonts if they're not currently embedded

leohemsted:
Review best served commit by commit

## Font embedding.

Essentially, PDFs have a setting contained within called `NeverEmbed`, which lists the fonts that it should never embed. By default, this is set to the 14 standard fonts (Times New Roman, Helvetica, Courier, Symbol, and Dingbats). When viewing a PDF this is fine because adobe reader will include those fonts - however, we have a hunch that the printing system that DVLA use doesn't have all these fonts installed on it. We can find unembedded fonts, and if so, embed them by setting `NeverEmbed` to an empty array within the PDF.

Note: The code that finds unembedded fonts finds more unembedded fonts than adobe acrobat. However we reckon a few false positives is fine.

Here's an example. The "before" file is a real letter that DVLA were unable to print. The "after" file is that file, having been run through the full sanitisation/validation/embedding/cmyk etc process.

| Before | After |
|--------|-------|
| image | image |

But don't just take my word for it, check out our customer testimonials!

> I can confirm that the attached letter ran successfully through to our Print queue.

Wow! You'd be a fool not to go for it. Dive, dive, dive!

## Refactoring

* Remove an unused `/convert.pdf` endpoint.
* Split up the `sanitise_file_contents` function, as flake8 complained it was too complex.
* Try and make all functions use BytesIO for consistency, because it was confusing working out what wanted bytes and what wanted a file-like.


app/precompiled.py
```diff
@@ -173,6 +164,25 @@ def sanitise_file_contents(encoded_string):
...
+def rewrite_pdf(file_data, page_count):
+    file_data, recipient_address, redaction_failed_message = rewrite_address_block(file_data)
+
+    if not recipient_address:
+        raise ValidationFailed("address-is-empty", [1], page_count=page_count)
+
+    if not does_pdf_contain_cmyk(file_data) or does_pdf_contain_rgb(file_data):
+        file_data = convert_pdf_to_cmyk(file_data)
+
+    if contains_unembedded_fonts(file_data):
```
servingUpAces: Do we want to log when this happens? Do we care how many time we have to embed fonts?

leohemsted: we log within the check function if it finds unembedded fonts

`current_app.logger.info(f'Found unembedded fonts {[x for x in unembedded]}')`

PR #420, opened 2020-02-24
redact pdfs with fonts that extend WinAnsi or MacRoman encoding

leohemsted:
Welcome to the exiting world of PDF fonts!

The PDF Redaction tool works in unicode, so that we can redact text that is unicode. The actual text is stored in binary within the pdf (as in a `b'string'`), and so pdf_redactor.py has to figure out what font that data is written in, and then decode it into unicode. Then, at the end it encodes it back into the original encoding.

Seems reasonable. However, pdf_redactor looks for a few things. The font needs a ToUnicode table, or to be one of WinAnsi or MacRoman encoding. If the font has none of these, then pdf_redactor just treats the text as a "?", so as a result it won't be redacted.

We have seen, in real letters, fonts that look like the following.

```python
{
    '/Type': '/Font',
    '/Subtype': '/TrueType',
    '/BaseFont': '/Arial',
    '/Encoding': {
        '/Type': '/Encoding',
        '/Differences': [
            '128',
            '/Euro'
        ],
        '/BaseEncoding': '/WinAnsiEncoding'
    },
    '/FirstChar': '30',
    '/LastChar': '255',
    '/Widths': (9, 0),
    '/FontDescriptor': (10, 0)
}
```

Here, rather than the /Encoding object being a string, it's a sub-object that lists a base encoding, and then a list of differences[1]. In the above example, it's WinAnsiEncoding, except for it replaces `chr(128)` with the euro symbol. (WinAnsiEncoding is `cp1252` in python). Now this is confusing, because WinAnsiEncoding already has the € symbol at 128[2], but who knows, I'm chalking this one up to arbitrary PDF bullshit as per usual.

Since this font is essentially identical to regular WinAnsi, I think it's safe to use the base encoding. Worst case: `string.encoding` crashes, we get a technical failure, an alert, and can investigate. My hunch is that that isn't especially common, but if it is, we could always work around it later.

refs:

[1] https://blog.idrsolutions.com/2011/04/understanding-the-pdf-file-format-%E2%80%93-custom-font-encodings/
[2] `chr(128) == '\x80'` and `b'\x80'.decode('cp1252') == '€'`


tests/test_precompiled.py
```diff
@@ -512,15 +513,15 @@ def test_is_notify_tag_calls_extract_with_wider_numbers(mocker):
     )
 
 
-def test_rewrite_address_block_end_to_end(mocker):
-    mock_escape = mocker.patch(
-        "app.precompiled.escape_special_characters_for_regex",
-        return_value='MR J DOE\n13 TEST LANE\nTESTINGTON\nTE57 1NG'
-    )
-    new_pdf, address, message = rewrite_address_block(BytesIO(example_dwp_pdf))
+@pytest.mark.parametrize(['pdf_data', 'address_snippet'], [
```
idavidmcdonald: Is the change to this test related to the change to the code above? Just can't quite put the two together

leohemsted: so this is taking the existing test, except parametrised to check both `example_dwp_pdf` and `valid_letter` - because `valid_letter` has mac fonts, while example_dwp_pdf has windows fonts. The valid_letter test case failed in my initial iteration of the code.

then i tweaked the assertions to make sure it returns the address too (previously it assigned the `address` variable but didn't check it.
i deleted the `escape_special_characters_for_regex` mock because it didn't feel appropriate for a test that specifically refers to being an end_to_end test.

idavidmcdonald: Ok, cheers. So it's not technically testing the new lines of code you added, it's more improving and adding coverage for an existing test, right?

If that's the case, is there any way to test the new functionality you've added regarding fonts with differences? If it's a pain to test you could maybe at the least add a simple unit test for `getEncoding` rather than trying to give `toUnicode` a PDF using a weird font

PR #427, opened 2020-03-05
Save letters which have failed redaction when sanitising

klssmith:
The new Celery tasks for sanitising letters were not going to be logging or copying letters that had failed the redaction of the address block - we've now decided that it would be worthwhile to copy them.

Previously, notifications-api was copying the letter to the `REDACTION_FAILURE` folder of the scan bucket. We've moved the copying to of the letter to template-preview instead since it saves having to pass extra data in the task argument argument to api. This does mean that we can't check if the letter was sent with a test key before copying it - we just copy all letters, but there is now a 7 day lifecycle on the bucket so we won't be storing the letters forever.


klssmith:
⚠️Don't review this yet - I forgot about a few bits... ⚠️

PR #436, opened 2020-03-24
Escape backslashes that are not newline/space characters in address regex

CrystalPea:
This way we avoid exception when doing letter redaction

Story ticket: https://www.pivotaltracker.com/story/show/171815617


app/precompiled.py
```diff
@@ -512,9 +512,14 @@ def _get_out_of_bounds_pages(src_pdf_bytes):
...
 
 def escape_special_characters_for_regex(string):
-    special_characters = ["[", "^", "$", ".", "|", "?", "*", "+", "(", ")"]
+    # those characters perform functions in regex expressions and have to be escaped:
+    special_characters = ["[", "{", "^", "$", ".", "|", "?", "*", "+", "(", ")"]
     for character in special_characters:
         string = string.replace(character, r"\{}".format(character))
+
+    bad_slashes = re.findall(r"\\[a-mo-rt-zA-Z]", string)
+    for s in list(set(bad_slashes)):
```
idavidmcdonald: Tests seem to pass without the `set`, I think you just need `list(bad_slashes)` here

tests/test_precompiled.py
```diff
@@ -593,13 +593,27 @@ def test_redact_precompiled_letter_address_block_sends_log_message_if_multiple_m
     assert "More than one match for address block during redaction procedure" in str(exc_info.value)
 
 
-def test_escape_special_characters_for_regex_matches_string():
-    string = 'PEA NUTT + MIKO JELLY4 JELLY COURT (PEANUT BUTTER JELLY WHARF)TOAST STR.ALLDAYSNACKSHIRESNACKISTANSN1 PBJ'
+@pytest.mark.parametrize('string', [
```
idavidmcdonald: Good work on the test cases with nice comments

app/precompiled.py
```diff
@@ -512,9 +512,14 @@ def _get_out_of_bounds_pages(src_pdf_bytes):
...
+        string = string.replace(s, "\\" + s)
```
leohemsted: i'm a bit nervous about this code. it looks pretty brittle. if someone passes in the string `\n` as part of a larger text , what does that read from the pdf? the string we get back from the pdf should be `b'\\n'` right?

what happens if someone passes in a pdf with a tab character (escaped that is `\t`) in the address? does that get picked up and changed into `\\t`?

i think we might want to rethink how this code works from the first principles and see

* what the string we get out of the extract address block function looks like
* what the regex we need for matching looks like

and then we can work out how best to transform that string to make it into that regex

app/precompiled.py
CrystalPea: Ahh, that's the case of missing test case :D. It is a set, because in a set values are unique, so we don't end up with too many backslashes :)

app/precompiled.py
CrystalPea: Hmmm, good point about the tabs. Maybe I could add "t" to omitted letters then, likes "s" and "n"? Right now our code throws an exception if there is "\P" or like in the address, so I think fixing that is more pressing than general improvements (which I'd love to do, but they will probably take longer and on my own I don't have ideas on how to do them).

tests/test_precompiled.py
CrystalPea: thanks :)

app/precompiled.py
CrystalPea: I added the missing test case now.

app/precompiled.py
CrystalPea:
```
if someone passes in the string \n as part of a larger text , what does that read from the pdf? the string we get back from the pdf should be b'\\n' right?
```
This PR would not change behaviour around that case.

leohemsted: my concern is I think that if someone passes through `\\t` that _isn't_ a tab character. That's a slash character `\` followed by a `t` character.

python only shows you `\t` as a convenience when you print strings. If there's a `\P` in our address, that's two characters, and I think we need to search for `\\P`.

if we just include `\` in the special characters does that work? 🤔 

i dont know this is giving me a headache thinking about it 😅

CrystalPea: > if we just include \ in the special characters does that work? 🤔

That would stop it from crashing, but it would not redact that address, because our regex would not be the same as the address in the document.

Hmm I still have the letter that broke it last time saved (I used it for testing that check), so if you have time next week we could go on a call and do some testing together, and maybe even come up with something better?

That is if it won't be another amazingly busy week ;)

CrystalPea: > my concern is I think that if someone passes through \\t that isn't a tab character. That's a slash character \ followed by a t character.

Ahh I see what you mean, you are right, we should definitely add a test case for what if the backslash is already escaped properly :o

leohemsted: yeah. if someone writes the text `bucki\ngham palace` in to their text editor, that'll become `bucki\\ngham` under the hood, and we should be preserving that. But the check will skip over it and the regex will end up being `bucki\s*gham` i think, which wouldn't match. aaaargh this is so confusing!

CrystalPea: Yeah this one is really tough, but at least it's not something introduced by this PR. I gather if we can make it better in one sitting, lets, if we can't, then just lets make sure nothing becomes worse, the exception gets fixed, deploy what we have, and then make a bigger overhaul plan.

PR #440, opened 2020-04-14
Address redaction fix

leohemsted:
# try fallback decoding before giving up redaction

for each font in the pdf file, if it has a ToUnicode map, then pdf_redactor tries to use that to convert the file to unicode. For each character, it uses the map to find the unicode translation. However, if for some reason the PDF has a character in a font, but doesn't have that character's code point in the map, then it'll replace the character with a "?". This isn't a ? that we see at the end - but this is a ? in the address that the regex can no longer match.

however, if the font doesn't have a ToUnicode map at all, then we try a couple of different fallback methods - we check the font's encoding, and if it's an encoding we know how to deal with (WinAnsiEncoding or MacRomanEncoding) then we decode the string that way.

For reasons unknown, this happened to 14000 PDFs uploaded last week. I can't recreate these PDFs without PII in them, so haven't been able to add a test case. But I've locally tested this against every single PDF that failed, and confirmed that all of those PDFs now rewrite their address properly.


app/pdf_redactor.py
```diff
@@ -551,16 +550,20 @@ def toUnicode(string, font, fontcache):
...
 
-        string = cmap.decode(string)
-        # print(string, end='', file=sys.stderr)
-        # sys.stderr.write(string)
-        return string
-    elif get_encoding(font) == "/WinAnsiEncoding":
+        try:
+            return cmap.decode(string)
+        except RedactionException:
+            # if there's a redaction exception, try and decode using fall back methods.
```
CrystalPea, line 556: do we still want to add a logging event here? Or do you think there is no need since we are rescuing it pretty cleanly? :)

scripts/run_app.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-gunicorn -w 5 -b 0.0.0.0:$1 wsgi
+flask run --host=0.0.0.0 -p $1
```
CrystalPea, line 2: This was a temporary change to use debugger on local, right?

leohemsted: I think we can keep this permanently. On paas we define the code that runs using the command var in the manifest, where we pass gunicorn through to run_app_paas. `exec scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py wsgi`

We run everything else locally through flask and through gunicorn on paas.

app/pdf_redactor.py
leohemsted: Yeah, that was my thinking. In practice this line executes for every text token, and in those rother PDFs we were often seeing four or five log lines.

We think it'll rescue cleanly, but if it doesn't:

* if the encoding lines fail (because it's got WinAnsiEncoding but it expects ToUnicode to work for some reason) and it for example instead encodes badly and returns gibberish, then we won't match, and will log `"No matches for address block during redaction procedure"`
* if it doesn't have Win or Mac encoding, then it'll log `f'Unrecognised font with encoding {font.Encoding}, may not be able to redact properly'`, return a `?`, and we'll log that No Matches log message again

so both are things we can pick up

PR #445, opened 2020-05-01
Allow international addresses in precompiled letters

quis:
Template preview gets precompiled letters from two places:
- from the admin app when someone uploads a PDF, to which it responds synchronously
- from one of the Celery queues, which it processes asynchronously

In both cases it checks the address on the letter to see if it’s valid, and returns an error message if it’s not. Now that some services have permission to send international letters, template preview needs to know whether it should allow an international address when it’s doing its validation. But template preview can’t just look at the service, because it doesn’t know about services.

Instead this commit adds optional arguments to both the Flask endpoint and Celery task that specify whether international addresses should be allowed. At a later date the admin app and API can start passing in these arguments based on what they know about the service.


app/precompiled.py
```diff
@@ -110,17 +110,23 @@ def rect(self, pt1, pt2):
 @statsd(namespace='template_preview')
 def sanitise_precompiled_letter():
     encoded_string = request.get_data()
+    allow_international_letters = (
+        request.args.get('allow_international_letters') == 'true'
```
klssmith, line 114: This is a `POST` endpoint only, so you won't be able to put `allow_international_letters` in `request.args`.

Also, given that api / admin will probably be passing through a boolean that has been converted to a string I guess we'd want to compare the value to `'True'` not `'true'` perhaps

quis: You can still `post` to a URL with with query string parameters. The alternative was to have two URLs, eg `/sanitise` and `/sanitise-international` but this felt less RESTful and would make the deployment order more rigid. We can’t put anything in the body of the request because it’s just the string encoded PDF, not JSON.

Putting capitalisation in URLs feels a bit wrong to me. I’d happy to make sure that the API/admin pass through the right string, rather than doing any type casting.

klssmith: 👍 I've not seen request args and POST mixed before, but I can see how it makes sense here

PR #447, opened 2020-05-04
Move print pdf onto a queue

CrystalPea:
In this PR:
- create a task to create pdfs for templated letters (based on `/print.pdf` endpoint)
- in this task upload the created pdf into S3 bucket from where we send files to DVLA
- call a task in API to set billing units for that letter

Deploy flow:
- first deploy this PR
- then deploy API PR
- then look what to do with `/print.pdf` endpoint (it might be used by other parts of Notify?)

Story ticket: https://www.pivotaltracker.com/story/show/171514311
API PR for the same story: https://github.com/alphagov/notifications-api/pull/2832


app/celery/tasks.py
```diff
@@ -84,3 +90,62 @@ def copy_redaction_failed_pdf(source_filename):
     # e.g. if a file has 5 days left to expire on a ONE_WEEK retention in the source bucket,
     # in the destination bucket the expiration time will be reset to 7 days left to expire
     obj.copy(copy_source, ExtraArgs={'ServerSideEncryption': 'AES256'})
+
+
+@notify_celery.task(name='create-letter-pdf')
+@statsd(namespace="template_preview")
+def create_letter_pdf(encrypted_letter_data):
```
servingUpAces, line 97: how about a name like `create-pdf-for-letter-template`? It makes it easier to distinguish between precompiled letter tasks and template letter tasks.

app/celery/tasks.py
```diff
@@ -84,3 +90,62 @@ def copy_redaction_failed_pdf(source_filename):
...
+
+    template = LetterPrintTemplate(
+        letter_details['template'],
+        values=letter_details['values'] or None,
+        contact_block=letter_details['letter_contact_block'],
+        # letter assets are hosted on s3
+        admin_base_url=current_app.config['LETTER_LOGO_URL'],
+        logo_file_name=logo_filename,
+    )
+    with current_app.test_request_context(''):
```
servingUpAces, line 110: why do we need to use the test_request_context here?

CrystalPea:
> Have you test this end to end with the API PR?

@servingUpAces I did end to end testing on local, both with and without API PR, I also checked AWS Amazon to see if the file is formed well and in the right bucket. All looked well.

app/celery/tasks.py
CrystalPea: ahh, that's a good idea, I will change that 👍

app/celery/tasks.py
CrystalPea: because instead of HTTP endpoint, our function is now a Celery task, and the HTML function it calls needs a request context that celery does not provide.

Or in other words, HTML function started throwing an exception about being outside of request context when I moved it onto a task, and I looked in weasyprint docs and found this little hack that made it work 😅

PR #470, opened 2020-07-31
reject letters with NOTIFY tags on later pages

leohemsted:
the NOTIFY tag signifies to dvla the first page of each letter - their print process combines our PDFs into one massive long one and then chops it up again based on those tags.

This means if there are NOTIFY tags on pages other than #1 they freak out cos it'll chop up into more envelopes than it was given files initially.

so change the code so that we use the existing text-checking function to check if NOTIFY is present on any of the later pages in the document. If it is, fail validation


app/precompiled.py
```diff
@@ -570,6 +574,25 @@ def rewrite_address_block(pdf, *, page_count, allow_international_letters):
 
 
 def _extract_text_from_pdf(pdf, *, x1, y1, x2, y2):
```
CrystalPea, line 576: shouldn't it be called `_extract_text_from_first_page_of_pdf` or something?

leohemsted: great point - squashed into the refactor commit

PR #471, opened 2020-07-31
Increase CF_STARTUP_TIMEOUT to 15 minutes

klssmith:
The app was taking over 10 minutes to deploy, so deploys were failing.


klssmith:
Now that we're only scaling template-preview to a max of 4 instances https://github.com/alphagov/notifications-paas-autoscaler/pull/100, this change shouldn't be needed

PR #480, opened 2020-09-10
Give a specific error code if a service tries to send international letters without permission

quis:
If a service doesn’t have permission to send international letters but someone tries to upload a letter with a valid international address we just tell them that the last line must be a UK postcode.

This is a bit opaque and:
- suggests that we’re not recognising at all that it’s not a UK letter
- doesn’t explain why it must be a UK postcode

This commit adds a new error code which will let us write a more helpful error message in this case.

***

Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/3621

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -105,6 +105,39 @@ class NotifyCanvas(canvas.Canvas):
         super().rect(left_x, bottom_y_from_bottom, width, height, fill=True, stroke=False)
 
 
+class PrecompiledPostalAddress(PostalAddress):
+
+    @property
+    def error_code(self):
+
+        if not self:
+            return "address-is-empty"
+
+        if not self.has_enough_lines:
+            return "not-enough-address-lines"
+
+        if self.has_too_many_lines:
+            return "too-many-address-lines"
+
+        if not self.has_valid_last_line:
+
+            if self.allow_international_letters:
+                return "not-a-real-uk-postcode-or-country"
+
+            if self.international:
+                return "cant-send-international-letters"
+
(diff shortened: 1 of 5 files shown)


app/precompiled.py
```diff
@@ -105,6 +105,39 @@ def rect(self, pt1, pt2):
         super().rect(left_x, bottom_y_from_bottom, width, height, fill=True, stroke=False)
 
 
+class PostalAddress(PostalAddress):
```
idavidmcdonald, line 108: Optional: I'm not sure if I like the inheriting a class with the same name. I feel it make things a bit opaque but I'm not decided

idavidmcdonald (commented):
Generally looks good but could we add a unit test please

app/precompiled.py
quis: I’ve renamed it to `PrecompiledPostalAddress` (since only precompiled letters need the address as a regex, for example).

tests/test_precompiled.py
```diff
@@ -516,6 +517,7 @@ def test_sanitise_precompiled_letter_with_missing_address_returns_400(client, au
 @pytest.mark.parametrize('file, allow_international, expected_error_message', (
     (bad_postcode, '', 'not-a-real-uk-postcode'),
     (bad_postcode, 'true', 'not-a-real-uk-postcode-or-country'),
+    (non_uk_address, '', 'cant-send-international-letters'),
```
quis, line 520: @idavidmcdonald I’ve added a unit test here

idavidmcdonald (approved):
Lovely work

PR #487, opened 2020-10-08
Fix the run_tests.sh script

klssmith:
**Fix the run_tests.sh script**
`run_tests.sh` is used to run the tests on Concourse. This had not been working as expected - unit tests were failing but the Concourse build was passing. This was because although the script was running the unit tests it was only calling the `display_result` function (which causes the script to exit if there are errors) after the code style checks. The `display_result` function should be called after each stage in the test script.

**Update cache keys used in tests**

**Update Dockerfile to use custom policy in test image**
We had made a change to the `policy.xml` file that ImageMagick uses. This change was only made in the production image, but the tests also need the updated `policy.xml` file too. This change moves the lines dealing with the file into the base image.

Also deleted a comment from the Dockerfile - the comment was added from this commit 3c2e8c9 but was now in the wrong place and no longer relevant.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -47,9 +47,10 @@ COPY docker/Arial.ttf /usr/share/fonts/truetype/msttcorefonts/
 
 WORKDIR /home/vcap/app
 
-# these are declared statically here so that they're cached by the docker image - if we run after the `COPY` command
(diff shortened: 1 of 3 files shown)


servingUpAces (approved):
Tests are passing locally

PR #492, opened 2020-10-27
Specify dockerhub username when deploying

idavidmcdonald:
The Dockerhub password will come from Concourse - Cloud Foundry looks for
the `CF_DOCKER_PASSWORD` environment variable. This means that we can
use our Dockerhub account to pull the Docker images and avoid any
problems with unauthenticated user rate limiting.

Equivalent to https://github.com/alphagov/notifications-antivirus/pull/60

https://www.pivotaltracker.com/story/show/175359312

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,8 @@ CF_API ?= api.cloud.service.gov.uk
 CF_ORG ?= govuk-notify
 CF_SPACE ?= development
 
-DOCKER_IMAGE = govuknotify/notifications-template-preview
+DOCKER_USER_NAME = govuknotify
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,7 +23,8 @@ CF_API ?= api.cloud.service.gov.uk
 CF_ORG ?= govuk-notify
 CF_SPACE ?= development
 
-DOCKER_IMAGE = govuknotify/notifications-template-preview
+DOCKER_USERNAME = govuknotify
```
klssmith, line 26: Just missing the `_` between `USER` and `NAME`

klssmith (approved):

PR #497, opened 2020-11-24
Upgrade Ghostscript to 9.53.3

klssmith:
This was tried [in October 2020](https://github.com/alphagov/notifications-template-preview/pull/486), then [reverted](https://github.com/alphagov/notifications-template-preview/pull/489) when it seemed like it might be causing DVLA problems. It now seems that the problems were unrelated, so they are happy for us to have another attempt at upgrading.

[Pivotal story](https://www.pivotaltracker.com/story/show/172077354)


idavidmcdonald (approved):
Lovely work.

To be merged on Wednesday please as DVLA would like to test this on the Thursday morning print run.

PR #501, opened 2020-12-08
Only redact first page

klssmith:
Previously, we were redacting the entire PDF, and then throwing away the
original pre-redaction data. however, we have issues that we haven't
been able to identify the root cause of, where the pdf redactor
sometimes throws away later pages of the pdf. An example was a ~10 page
PDF where the redactor only returned the first three pages.

We know we only want to redact the first page, so we can just pass that
first page in to the pdf_redactor code. Then, we need to combine that
new redacted first page with the existing pages from the user's input
pdf, so a new function `replace_first_page_of_pdf_with_new_content` has
been written that just stitches the new page on top of the remaining
pages from the original data.

[Pivotal story](https://www.pivotaltracker.com/story/show/175989270)


CrystalPea (approved):
Stellar work 🌟

PR #515, opened 2021-02-15
Retry if creating letter PDF transiently fails

benthorner:
https://www.pivotaltracker.com/story/show/176804965

Please see the commit message for more details.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,6 +1,8 @@
 import base64
 from io import BytesIO
 
+import werkzeug.exceptions
+
 from botocore.exceptions import ClientError as BotoClientError
 from flask import current_app
 from flask_weasyprint import HTML
@@ -95,9 +97,9 @@ def copy_redaction_failed_pdf(source_filename):
     obj.copy(copy_source, ExtraArgs={'ServerSideEncryption': 'AES256'})
 
 
-@notify_celery.task(name='create-pdf-for-templated-letter')
+@notify_celery.task(bind=True, name='create-pdf-for-templated-letter', max_retries=3, default_retry_delay=180)
 @statsd(namespace="template_preview")
-def create_pdf_for_templated_letter(encrypted_letter_data):
+def create_pdf_for_templated_letter(self, encrypted_letter_data):
     letter_details = current_app.encryption_client.decrypt(encrypted_letter_data)
     current_app.logger.info(f"Creating a pdf for notification with id {letter_details['notification_id']}")
     logo_filename = f'{letter_details["logo_filename"]}.svg' if letter_details['logo_filename'] else None
@@ -113,7 +115,10 @@ def create_pdf_for_templated_letter(encrypted_letter_data):
     with current_app.test_request_context(''):
         html = HTML(string=str(template))
 
-    pdf = BytesIO(html.write_pdf())
+    try:
+        pdf = BytesIO(html.write_pdf())
+    except werkzeug.exceptions.BadGateway as exc:
+        self.retry(exc=exc, queue=QueueNames.SANITISE_LETTERS)
 
     cmyk_pdf = convert_pdf_to_cmyk(pdf)
     page_count = get_page_count(cmyk_pdf.read())
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,8 +1,6 @@
 import base64
 from io import BytesIO
 
-import werkzeug.exceptions
-
 from botocore.exceptions import ClientError as BotoClientError
 from flask import current_app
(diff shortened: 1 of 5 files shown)


app/celery/tasks.py
```diff
@@ -95,9 +97,9 @@ def copy_redaction_failed_pdf(source_filename):
     obj.copy(copy_source, ExtraArgs={'ServerSideEncryption': 'AES256'})
 
 
-@notify_celery.task(name='create-pdf-for-templated-letter')
+@notify_celery.task(bind=True, name='create-pdf-for-templated-letter')
```
klssmith, line 100: I think the default number of retries and the default delay between retries are good, but it might be useful to add them in as arguments anyway (even if you keep the defaults). It would help to make it obvious what they were and mean that we don't get caught out if the defaults change in future.

app/celery/tasks.py
```diff
@@ -113,7 +115,10 @@ def create_pdf_for_templated_letter(encrypted_letter_data):
     with current_app.test_request_context(''):
         html = HTML(string=str(template))
 
-    pdf = BytesIO(html.write_pdf())
+    try:
+        pdf = BytesIO(html.write_pdf())
+    except werkzeug.exceptions.BadGateway as exc:
+        self.retry(exc=exc)
```
klssmith, line 121: I think you need to add `queue` here as a kwarg (I looked in the docs and couldn't see which queue would be used if that's not specified).

app/celery/tasks.py
```diff
@@ -113,7 +115,10 @@ def create_pdf_for_templated_letter(encrypted_letter_data):
     with current_app.test_request_context(''):
         html = HTML(string=str(template))
 
-    pdf = BytesIO(html.write_pdf())
+    try:
+        pdf = BytesIO(html.write_pdf())
+    except werkzeug.exceptions.BadGateway as exc:
```
klssmith, line 120: When this happened a few days ago I could only see a `weasyprint.urls.URLFetchingError` exception and not a `BadGateway` exception. How about just catching `Exception` here to cover anything going wrong?

Optional - a log line to tell us what type of exception was raised might then be useful

app/celery/tasks.py
benthorner: Sure, I've done that 👍.

app/celery/tasks.py
benthorner: We paired on this and confirmed the `queue` kwarg is necessary, which is strange because [the docs](https://docs.celeryproject.org/en/3.1/userguide/tasks.html#retrying) for v3 celery explicitly say: "it will take care to make sure the message is delivered to the same queue as the originating task", but the code doesn't clearly match up to that ([here](https://github.com/celery/celery/blob/6cf4f40fc58fe8585721f4df95a1c77d25106dbf/celery/app/task.py#L600)). There's also [this issue](https://github.com/celery/celery/issues/3622), suggesting a similar problem.

Anyway, I've amended the commit and specified the queue as "SANITISE_LETTERS", which is the one the API app creates instances of this task on ([here](https://github.com/alphagov/notifications-api/blob/master/app/celery/letters_pdf_tasks.py#L89)). We _could_ have a special "RETRY" queue like the API uses, but that seems overkill for something that we expect to be infrequent and for which we only retry for a short period.

(Also tested it works as expected.)

app/celery/tasks.py
benthorner: The BadGateway is the top-level exception. The one you mentioned would be clearer, but it's not the one that we get, unfortunately. Here's an example stack trace:

```
...

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
...
  File "/usr/local/lib/python3.6/site-packages/weasyprint/urls.py", line 286, in fetch
    raise URLFetchingError('%s: %s' % (type(exc).__name__, str(exc)))
weasyprint.urls.URLFetchingError: URLError: <urlopen error _ssl.c:835: The handshake operation timed out>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
...
  File "/usr/local/lib/python3.6/site-packages/werkzeug/exceptions.py", line 807, in __call__
    raise self.mapping[code](*args, **kwargs)
werkzeug.exceptions.BadGateway: 502 Bad Gateway: The proxy server received an invalid response from an upstream server.
```

Looking again at the stack trace, I realised there was something strange going on. I've added another commit to explain and refactor the strangeness so it's a bit less strange.

Could you take a look?

klssmith: 👍 that makes sense - I didn't spot the Bad Gateway error at first because I was looking at a different error log (the one for `Failed to load image at...` not for the task as a whole)

klssmith (approved):

PR #517, opened 2021-02-25
Simplify README and consolidate scripts

benthorner:
Partly addresses: https://github.com/alphagov/notifications-manuals/issues/9

Please see the commit messages for more details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,9 +62,8 @@ clean: ## Clean up any remaining docker containers
 .PHONY: upload-to-dockerhub
 upload-to-dockerhub:
 	docker build -f docker/Dockerfile -t ${DOCKER_IMAGE_NAME} .
-	$(if ${DOCKERHUB_USERNAME},,$(error Must specify DOCKERHUB_USERNAME))
 	$(if ${DOCKERHUB_PASSWORD},,$(error Must specify DOCKERHUB_PASSWORD))
-	@docker login -u ${DOCKERHUB_USERNAME} -p ${DOCKERHUB_PASSWORD}
+	@docker login -u govuknotify -p ${DOCKERHUB_PASSWORD}
 	docker push ${DOCKER_IMAGE_NAME}
 
 # ---- PAAS COMMANDS ---- #
--- a/Makefile
+++ b/Makefile
@@ -6,19 +6,11 @@ APP_VERSION_FILE = app/version.py
 
 GIT_COMMIT ?= $(shell git rev-parse HEAD)
 
-TEMPLATE_PREVIEW_INTERNAL_SECRETS ?= '["my-secret-key"]'
-
-DOCKER_CONTAINER_PREFIX = ${USER}-notifications-template-preview-manual
-
 NOTIFY_CREDENTIALS ?= ~/.notify-credentials
 
-NOTIFY_APP_NAME ?= notify-template-preview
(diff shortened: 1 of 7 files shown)


leohemsted (commented):
Looks pretty good, changes to docker process sounds great too. Just one small comment on the deploying section, think it might benefit from a bit more detail

Makefile
```diff
@@ -35,114 +27,62 @@ PORT ?= 6013
...
+.PHONY: bootstrap
+bootstrap: _generate-version-file
+	mkdir -p log # manually create directory to avoid permission issues
+	docker build -f docker/Dockerfile --target test -t notifications-template-preview .
 
-.PHONY: bash-with-docker
-bash-with-docker: prepare-docker-test-build-image ## Build inside a Docker container
-	$(call run_docker_container,build, bash)
+.PHONY: run-flask
+run-flask:
```
servingUpAces, line 45: Can we add the help comments (adding ## run flask in docker container), please?

README.md
```diff
@@ -59,17 +83,11 @@ curl \
...
-```shell
-make run-celery-with-docker
-```
-
 ## Deploying
 
-You shouldn’t need to deploy this manually because there’s a pipeline setup in Concourse. If you do want to deploy it manually, you'll need the notify-credentials repo set up locally. `CF_APP` should be set to `NOTIFY_TEMPLATE_PREVIEW_CELERY` if deploying the Celery app.
+You shouldn’t need to deploy this manually because there’s a pipeline setup in Concourse.
+
+If you do want to deploy it manually, you'll need the notify-credentials repo set up locally. `CF_APP` should be set to `NOTIFY_TEMPLATE_PREVIEW_CELERY` if deploying the Celery app.
```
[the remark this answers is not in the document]
benthorner: Sure, that's a fair point. I've added another commit with the revised instructions. Could you take a look?

On a side note, I noticed the deploy tasks are also slightly out-of-sync between Concourse and the Makefile - the Concourse one doesn't use `upload-to-docker`. Ideally Concourse would rely on the Makefile, so there's a stronger guarantee the tasks will continue working. I might address that when I come to [the second phase](https://github.com/alphagov/notifications-manuals/issues/9) of the issue.

I agree it

Makefile
benthorner: Ahh...after reading your comment I realised these are used for `make help`. Sure, I've amended the 5th commit with some comments. Does that look OK now?

leohemsted (commented):
just uncovered something in the dockerhub creds that i'd like to understand a bit more 😵

leohemsted (approved):
nice

README.md
```diff
@@ -1,41 +1,65 @@
...
 
-Running tests will also apply syntax checking, using [flake8](https://pypi.org/project/flake8/).
+## To test the application
 
+```shell
+make test
+```
 
-## Running the Flask application
+If you need to run a specific command, such as a single test, you can use the `run_with_docker.sh` script. This is what `test-with-docker` and other `make` rules use.
```
CrystalPea, line 37: Hey, how do we use `run_with_docker.sh` to run a single test?

benthorner: There's an example of doing that directly below 😄.

CrystalPea: Ahh, yes, you are right, sorry 🤦🏼‍♀️

PR #520, opened 2021-03-09
Add logging to test the letters before adding code to fail the letters

servingUpAces:
Add logging when the PDF still contains unembedded fonts after the remove function has run.
I'd like to mark the letters as validation failed for the letters that
do, but without real letters to test with I am worried it might fail
good letters. This is just an extra precaution.

Once I'm happy with the test, if the letter still contains unembedded
fonts after the remove function mark the letter as validation failed.
But what should the failure message look like?


leohemsted (approved):

CrystalPea (approved):
Looks good.

Maybe in the future we could pass notification id to template preview from admin/api along with the file? That would make debugging much easier 🤔

PR #523, opened 2021-03-19
add FreeSans as a fallback font

leohemsted:
We use Nimbus Sans L for our templated letter font, however, some characters aren't included in that font (notably Punjabi). install the freefont package so that utils can render those characters using that character set.

The font is used in the css added here: https://github.com/alphagov/notifications-utils/pull/852

interestingly it appears to have modified the font for al

| Before      | After |
| ----------- | ----------- |
| image | image |
| image   | image        |

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,6 +25,7 @@ RUN echo "Install base packages" && apt-get update \
         gsfonts \
         libcurl4-openssl-dev \
         libssl-dev \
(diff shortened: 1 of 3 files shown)


idavidmcdonald (approved):
This is an improvement but as your screenshots point out, doesn't solve this for all languages. Could be worth trying to add further support in time for that

servingUpAces (approved):

PR #526, opened 2021-03-29
Clean up a few more Makefile rules

benthorner:
These were missed in [1] and identified in [2] (see last two commits).

[1]: https://github.com/alphagov/notifications-template-preview/pull/517
[2]: https://github.com/alphagov/notifications-antivirus/pull/67

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -29,5 +29,8 @@ fi
 flake8 .
 display_result $? 1 "Code style check"
 
+isort --check-only ./app ./tests
+display_result $? 2 "Import order check"
+
 py.test --maxfail=10 tests/ --strict
-display_result $? 2 "Unit tests"
+display_result $? 3 "Unit tests"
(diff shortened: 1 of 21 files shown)


scripts/run_tests.sh
```diff
@@ -29,5 +29,8 @@ fi
 flake8 .
 display_result $? 1 "Code style check"
 
+isort --check-only ./app ./tests
+display_result $? 2 "Import order check"
+
 py.test --maxfail=10 tests/ --strict
 display_result $? 2 "Unit tests"
```
klssmith, line 36: Should this be `3` not `2` now to give each failure a unique exit status? Probably doesn't matter too much - I'm not sure that we use the exit statuses anywhere

klssmith (approved):

scripts/run_tests.sh
benthorner: Oh yes, good spot. I hadn't noticed the return codes, and don't think they're used. Nonetheless, I've changed it to a `1 2 3` sequence for now.

PR #531, opened 2021-04-13
Revamp Celery stats and request tracing

benthorner:
https://www.pivotaltracker.com/story/show/176261229

This applies the same change we made in the API, and goes a bit further
to enable end-to-end request tracing, which is part of the same code.

Please see the commit messages for more details. If we have concerns about
the request tracing part, I can split that into a separate PR.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -105,8 +105,6 @@ def create_app():
 
     load_config(application)
 
-    notify_celery.init_app(application)
-
     from app.precompiled import precompiled_blueprint
     from app.preview import preview_blueprint
     from app.status import status_blueprint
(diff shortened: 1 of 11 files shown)


leohemsted (approved):
seems fine. a shame we had to remove the autouse, but it's not such a big hassle to add `client` to params for tests so thats okay

benthorner:
Just a quick note before this goes in about request tracing: this PR is only concerned with getting it working at the Celery level. Request tracing at the web level will require separate changes to [the Admin app](https://github.com/alphagov/notifications-admin/blob/007fa5fa19d810d04df1c102c37837feb8cda80b/app/template_previews.py) to inject the `X-B3-TraceId` header.

PR #542, opened 2021-06-04
Move original precompiled PDFs to backup S3 bucket

CrystalPea:
Those letters will be kept there for a week. We are doing this so that if we push risky updates to dependencies or there are any other problems with letter processing, or if we need to investigate why some letters cause trouble to DVLA, we have the originals as backup to either re-send or investigate any issues.

### DIFF
--- a/tests/celery/test_tasks.py
+++ b/tests/celery/test_tasks.py
@@ -33,11 +33,21 @@ def test_sanitise_and_upload_valid_letter(mocker, client):
 
     sanitise_and_upload_letter('abc-123', 'filename.pdf')
 
-    mock_upload.assert_called_once_with(
-        filedata=mocker.ANY,
-        region=current_app.config['AWS_REGION'],
-        bucket_name=current_app.config['SANITISED_LETTER_BUCKET_NAME'],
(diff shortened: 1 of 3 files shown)


tests/celery/test_tasks.py
```diff
@@ -33,11 +33,21 @@ def test_sanitise_and_upload_valid_letter(mocker, client):
...
+    mock_upload.assert_has_calls(
+        [
+            call(
+                filedata=mocker.ANY,
+                region=current_app.config['AWS_REGION'],
+                bucket_name=current_app.config['SANITISED_LETTER_BUCKET_NAME'],
+                file_location='filename.pdf',
+            ),
+            call(
+                filedata=mocker.ANY,
```
leohemsted, line 45: are we able to replace this mocker.ANY call with `valid_file`?

leohemsted (approved):

tests/celery/test_tasks.py
CrystalPea: lets see :3

CrystalPea: done ☺️

PR #552, opened 2021-06-23
Add missing tests and comments for CMYK conversion

benthorner:

This is one of the ways we use GhostScript, which we want to upgrade.
Before we do that, we should ensure it's thoroughly tested.

Please see the commits for more details.


benthorner:
Interestingly, the CMYK "true" black looks less black on my monitor than the pseudo black. Apparently my monitor isn't a very good CMYK printer! Can YOU tell the difference? 😄 

| Pseudo Black ([184, 173, 171, 226])  | True Black ([0, 0, 0, 255]) |
| Screenshot 2021-06-23 at 13 12 53 | Screenshot 2021-06-23 at 13 12 46 |

servingUpAces (approved):

PR #553, opened 2021-06-23
Fix not detecting Type 0 (composite) fonts

benthorner:

Previously the function to gather used fonts would only walk certain
parts of the PDF object tree. This worked for PDFs which had TrueType
fonts embedded, since the tree looks like this:

```
(parents)
  BaseFont: AAABQG+ArialMT
  ...
  FontDescriptor:
    FontFile2: ...
    FontName: AAABQG+ArialMT
    ...
```

Some PDFs have composite fonts, which look like this:

```
(parents)
  BaseFont: MUFUZY+ArialMT
  ...
  DescendantFonts: # AN ARRAY!
    - BaseFont: MUFUZY+ArialMT
      FontDescriptor:
        FontFile2: ...
        FontName: MUFUZY+ArialMT
        ...
```

In addition, the FontDescriptor is actually an indirect reference,
which means it doesn't (directly) have a "keys" attribute.

This fixes the false positives for detecting unembedded fonts, by
adding support to walk the tree for arrays and indirect objects.

Note, I found this article helpful [1].

[1]: https://www.prepressure.com/pdf/basics/fonts

### DIFF
--- a/app/embedded_fonts.py
+++ b/app/embedded_fonts.py
@@ -1,11 +1,12 @@
 import subprocess
 from io import BytesIO
 
+from flask import current_app
 from PyPDF2 import PdfFileReader
 from PyPDF2.generic import IndirectObject
 
 
-def contains_unembedded_fonts(pdf_data):  # noqa: C901 (too complex)
+def contains_unembedded_fonts(pdf_data, filename=''):  # noqa: C901 (too complex)
     """
     Code adapted from https://gist.github.com/tiarno/8a2995e70cee42f01e79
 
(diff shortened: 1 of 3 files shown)


benthorner:
(I used PDFBox PDFDebugger to look at the PDF object tree.)

benthorner:
I've added one extra commit to do some logging (it's in the same code, so may as well do it in this PR).

idavidmcdonald (approved):
Well done, some good work on some of our most complex code

A few suggestions

- you may like to contribute the fix in the first commit into the gist that we originally took this code from
- be careful of the slippery slope of trying to fix the problems DVLA had had with the occasional letter if you are aiming to instead be focussing on the ghostscript upgrade. Saying that, I wouldn't be surprised if you are the person who finally comes along and cracks it but remember not to go to deep into that rabbit hole for now 

Also I think the logging is useful for type 3 fonts. I don't know whether the correct approach will be to reject letters using a type 3 font as this may not be good for our users. Imagine if you got told your letter can't be accepted because of a type 3 font... how on earth do you know how to fix it? But that is a conversation for another day that this logging will help us have

benthorner:
Thanks!

> you may like to contribute the fix in the first commit into the gist that we originally took this code from

That's a nice suggestion. [I've done that](https://gist.github.com/tiarno/8a2995e70cee42f01e79#gistcomment-3792657).

> be careful of the slippery slope of trying to fix the problems DVLA had

Indeed. It made sense to add the logging while I'm here, but it's far from clear if Type3 is really the issue, and we'd need to agree how to reject them in a way that's understandable.

PR #554, opened 2021-06-25
Remove unnecessary re-check of unembedded fonts

benthorner:
Follows on from: https://github.com/alphagov/notifications-template-preview/pull/553

Since we fixed the method to detect these, we've not seen any logs
about them, which indicates it's not an issue. We can always put
the log back if we believe it might be.


benthorner:
We haven't seen any of these logs for over a week and yet PDFs have been failing to print. 

I think it's OK to remove them, as they're clearly not a reliable indicator of issues with PDFs.

klssmith (approved):

PR #556, opened 2021-06-30
Make bootstrap consistent with Ruby API Client

benthorner:
It's taken a few attempts to standardise the way we do local setup
for repos that run through Docker. The final repos to standardise
were the API clients, where we agreed on a slightly different way
of doing local setup [1]. This change is to be consistent with that.

Note that the Dockerfile already had a "COPY . ." statement, which
supersedes "COPY requirements_for_test.txt ." since it copies all
files. Since we no longer need to rebuild the image every time we
run the tests, the previous comment [2] no longer applies.

[1]: https://github.com/alphagov/notifications-ruby-client/pull/122
[2]: https://github.com/alphagov/notifications-template-preview/blob/f9df02478683896a38b66f4f77d50775729905da/docker/Dockerfile#L66

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+environment*.sh
--- a/Makefile
+++ b/Makefile
@@ -34,11 +34,15 @@ help:
 generate-version-file:
 	@echo -e "__commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"" > ${APP_VERSION_FILE}
 
+.PHONY: bootstrap
+bootstrap: generate-version-file
+	mkdir -p log # manually create directory to avoid permission issues
+	pip install -r requirements_for_test.txt
+
 # ---- DOCKER COMMANDS ---- #
 
 .PHONY: bootstrap
-bootstrap: generate-version-file ## Setup environment to run app commands
-	mkdir -p log # manually create directory to avoid permission issues
+bootstrap-with-docker: ## Setup environment to run app commands
 	docker build -f docker/Dockerfile --target test -t notifications-template-preview .
 
 .PHONY: run-flask-with-docker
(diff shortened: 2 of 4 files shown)


leohemsted (commented):
do we still want `copy . .` at all? we're mounting the current directory as a volume in `/home/vcap/app` that will just lead to confusion when the docker container will potentially have outdated versions of files in that bear no resemblance to what is actually in it when you run the tests and it mounts the current working directory. I've had this as a gotcha recently when working with docker containers locally.

i'd be interested in thinking about copying just the reqs.txt, makefile, etc, before running `make bootstrap`. But not copying the other files or folders so that it's clear if someone starts up a new container that they won't have any source code, just the environment

benthorner:
@leohemsted ideally yes, the image would be minimal. A minor point is that it's just a bit more faff to identify the files that are necessary for the subsequent commands - ideally we'd remove them afterwards as well. The major point is that CI needs the full set of files, since [it uses the vanilla image to run the tests](https://github.com/alphagov/notifications-aws/blob/b246fb5106c65eec5e2e16ccddbdc0c093989671/concourse/templates/macros.yml.j2#L85-L87), without mounting over the top of the files.

> I've had this as a gotcha recently when working with docker containers locally.

I haven't experienced this myself (except when making these changes!). As long as you use the `run_with_docker.sh` script, then in theory you should never see the out-of-date files in the image.

Does that help a bit?

leohemsted (approved):
Good point about CI. `copy . .` can stay 👍

benthorner:
Thanks @leohemsted!

As discussed separately, although neither of us are concerned about secrets making their way into images - especially since the only "secret" for this repo is the SQS prefix! - I decided to make this consistent with [other repos](https://github.com/alphagov/notifications-ruby-client/pull/124) and avoid it as a good practice. I've added a `.dockerignore` in a second commit. That look OK to you?

In case you're feeling energetic: now we're happy with the approach here, I've raised [an identical PR for antivirus](https://github.com/alphagov/notifications-antivirus/pull/74).

PR #558, opened 2021-07-01
Log info about precompiled letter software

benthorner:

See the comment for more details. If you're wondering what the difference 
between "creator" and "producer" is:

- The "creator" is the original publishing software.
- The "producer" is the software used to convert the original format to a PDF (if any).

### DIFF
--- a/tests/test_precompiled.py
+++ b/tests/test_precompiled.py
@@ -27,6 +27,7 @@ from app.precompiled import (
     get_invalid_pages_with_message,
     handle_irregular_whitespace_characters,
     is_notify_tag_present,
+    log_metadata_for_letter,
     redact_precompiled_letter_address_block,
     replace_first_page_of_pdf_with_new_content,
     rewrite_address_block,
@@ -51,6 +52,7 @@ from tests.pdf_consts import (
     non_uk_address,
     not_pdf,
     notify_tags_on_page_2_and_4,
+    pdf_with_no_metadata,
     portrait_rotated_page,
     repeated_address_block,
     sample_pages,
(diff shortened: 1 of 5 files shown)


CrystalPea (approved):

benthorner:
> are they ever not present? (and will they ever AttributeError?)

@leohemsted raised a good point. Although the individual attributes will be fine, it's also possible for a PDF not to have an `/Info` section. I've added a new commit to handle that, with a test to cover it properly.

@CrystalPea do you want to take another look?

tests/test_precompiled.py
```diff
@@ -558,6 +561,20 @@ def test_sanitise_precompiled_letter_with_bad_address_returns_400(
...
+    pdf_with_no_metadata,
+], ids=['valid_letter', 'pdf_with_no_metadata'])
+def test_log_metadata_for_letter(
+    client,
+    file,
+    mocker,
+):
+    logger = mocker.patch('app.precompiled.current_app.logger.info')
+    log_metadata_for_letter(BytesIO(file), 'filename')
+    assert logger.called
```
CrystalPea, line 575: would we like to assert that it's called with a correct message?

benthorner: That would be more effort, but the log isn't critical: just for analysis.

I've tested it works manually, so I'm satisfied it's been sufficiently tested for the time being.

If it does break in future, we can always make the test more prescriptive.

CrystalPea: ok 👍🏼

CrystalPea (approved):

PR #582, opened 2021-10-29
Update to Python 3.9

rjbaker:
Updates base container image to Python 3.9 based on Debian Buster.

Updates PyMuPDF to a later version. The previous version was fairly old and had no wheel distributions compatible with Python 3.9.

Updates deprecated method names in PyMuPDF fitz module. These were simply changed to `snake_case`.

Add `#pyup ignore` to various dependencies flagged on this PR. These do not block the upgrade to Python 3.9 and will be tackled in separate PRs. Various refactors are likely and out of scope of this update. Discussed with @quis.

🚨  **Accurate PDF rendering is untested. Don't merge until this has been checked by 👁️ **

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,18 +1,18 @@
 #app requirements
 boto3==1.18.6
-celery[sqs]==5.0.5
+celery[sqs]==5.0.5 # pyup: ignore
 Flask==1.1.4
 Flask-WeasyPrint==0.6
 Flask-HTTPAuth==4.2.0
-Werkzeug==1.0.1
+Werkzeug==1.0.1 # pyup: ignore 
 
 # pdf libraries
 html5lib==1.1
 wand==0.5.9
 jsonschema==3.2.0
 pypdf2==1.26.0
-Pillow==7.2.0
-reportlab==3.5.67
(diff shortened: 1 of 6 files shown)


leohemsted (approved):
code changes look reasonable, but i haven't tested manually for visual comparison of output or anything

requirements.txt
```diff
@@ -1,20 +1,20 @@
...
 wand==0.5.9
 jsonschema==3.2.0
 pypdf2==1.26.0
-Pillow==7.2.0
-reportlab==3.5.67
+Pillow==7.2.0 # pyup: ignore
+reportlab==3.5.67 # pyup: ignore
 pdf2image==1.12.1
-PyMuPDF==1.16.11
+PyMuPDF==1.19.1
```
quis, line 17: I think we only use this (or rather the `fitz` package which it ships) to read from PDFs.

So if the existing tests to read from PDFs pass then I’m pretty confident this doesn’t need a visual check.

CrystalPea (approved):
LGTM 👍🏼 

Should we deploy it on a Monday afternoon, around 3-4pm, so if any letters are affected, it's not that many and DVLA would tell us the very next day?

PR #584, opened 2021-11-08
Add tests for PDF boundaries

klssmith:
This adds two new test PDFs. One has all writable areas filled in blue - with this one we test that no content is over the boundaries. The other is a one page PDF filled with text in all writable areas - we assert that this is valid and only one page in case a future change to the code affects how the PDFs are written.


tests/test_precompiled.py
```diff
@@ -438,6 +439,29 @@ def test_precompiled_sanitise_pdf_without_notify_tag(client, auth_header):
     )
 
 
+def test_precompiled_sanitise_pdf_for_pdf_with_content_near_margins(client, auth_header):
```
benthorner, line 442: This test may not flag a bunch of issues:

- We [perform validation / count pages](https://github.com/alphagov/notifications-template-preview/blob/03df0d92522f13091b081f3fe04c188e85d2ade6/app/precompiled.py#L174-L179) before rewriting, so if the rewrite code trashes the PDF we may not know, as the error could still be the same.

- Some of [the rewriting code](https://github.com/alphagov/notifications-template-preview/blob/03df0d92522f13091b081f3fe04c188e85d2ade6/app/precompiled.py#L224-L228) won't run on this sample file, as it doesn't need CMYK or unembedded font conversion (I manually checked the predicates).

So I'm not sure we get much from this test. I was thinking about how to address the overall concern of "something might trash the PDF and we wouldn't know". We could use a different technique: output fixtures. This would involve passing a PDF - maybe the example DWP one - through the code and saving the result, so that we can check the bytes are identical each time. Candidate functions:

- `rewrite_address_block`
- `convert_pdf_to_cmyk`
- `embed_fonts`
- `add_notify_tag_to_letter`

As long as we make it clear how to regenerate the output fixtures (and do a visual check they look OK), then these tests could provide a lot more reassurance beyond the text-snippet-oriented ones we have at the moment.

What do you think?

benthorner (commented):
Great digging with the first test ⭐. I'm not so sure about the second one, though and have left a suggestion for that. How did you find the gap in the tests? It's quite a big codebase to go through so really appreciate the effort here!

I had a look through myself and found a couple of other places that might be worth having some extra tests for. Interested to hear your thoughts on these:

- [`_colour_no_print_areas_of_page_in_red`](https://github.com/alphagov/notifications-template-preview/blob/03df0d92522f13091b081f3fe04c188e85d2ade6/app/precompiled.py#L501) - this is called from several places but we never actually test its behaviour (all calls to it are mocked).

- [`replace_first_page_of_pdf_with_new_content`](https://github.com/alphagov/notifications-template-preview/blob/03df0d92522f13091b081f3fe04c188e85d2ade6/app/precompiled.py#L827) - we do [test this replaces stuff and maintains the page count](https://github.com/alphagov/notifications-template-preview/blob/03df0d92522f13091b081f3fe04c188e85d2ade6/tests/test_precompiled.py#L778-L779), but not if the content on subsequent pages is unchanged.

I don't think the first one is particularly critical as it doesn't affect what we send. It could mean we start showing weird stuff to users if they have invalid PDFs. The second case is just a couple of missing assertions, or it could also be covered as part of the "output fixture" technique I mentioned in my comment on your second test.

klssmith:
Closing in favour of [this story](https://www.pivotaltracker.com/story/show/180261057). I'm not cherry-picking the first commit out because that was the test I wasn't entirely convinced about - I was more keen on the second test. I'll let whoever picks up the new story decide if they want to re-use any of this code or not.

PR #588, opened 2021-11-15
Pin all dependencies using pip-tools

idavidmcdonald:
We do this because we want to make sure everytime we deploy
this app, we aren't changing any dependencies unless we explicitely
want to.

This follows the approach of the admin app which uses pip-tools.

I only had to change one top level dependency, boto3 from 1.18.6
to 1.20.5 as pip-tools was complaining that it could not resolve
a version of botocore due to conflicting requirements. Bringing
boto3 up to the latest available version seems to fix this.


Makefile
```diff
@@ -23,13 +23,20 @@ DOCKER_IMAGE_NAME = ${DOCKER_IMAGE}:${DOCKER_IMAGE_TAG}
...
 .PHONY: help
 help:
 	@cat $(MAKEFILE_LIST) | grep -E '^[a-zA-Z_-]+:.*?## .*$$' | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 # ---- LOCAL FUNCTIONS ---- #
 # should only call these from inside docker or this makefile
 
+.PHONY: freeze-requirements
+freeze-requirements: ## create static requirements.txt
+	${VIRTUALENV_ROOT}/bin/pip install --upgrade pip-tools
```
benthorner, line 37: Do you think it's worth supporting `venv`? We don't [here](https://github.com/alphagov/notifications-api/blob/master/Makefile#L64), so having sporadic support among our apps is likely low value / confusing.

Also notable we run bare `pip` below, so if we do this here we should make it consistent elsewhere in this file.

benthorner (commented):
Nice, definitely prefer this tool 🚀.

I'm sure it's safe to upgrade `boto` for this repo. I've left an almost optional (but not quite, sorry!) comment where I noticed a different between this app and others.

Makefile
idavidmcdonald: I'm OK to get rid of it and just use bare `pip` and assume that people are using a correct version provided by pyenv for example. I don't think it is risk free, as people are maybe a bit less likely to be in a virtualenv as they instead just rely on things happening in docker but potential impact is low.

Let's see how we get on!

idavidmcdonald: Have pushed up with a rebased commit

Makefile
```diff
@@ -23,13 +23,19 @@ DOCKER_IMAGE_NAME = ${DOCKER_IMAGE}:${DOCKER_IMAGE_TAG}
 
 PORT ?= 6013
 
+
```
benthorner, line 26: 😺

benthorner (approved):
Tested and works locally with `./scripts/run_with_docker.sh make freeze-requirements` ✅ 

Regarding [the use of venv](https://github.com/alphagov/notifications-template-preview/pull/588#discussion_r749514261), it's worth noting the original version didn't work in Docker:

```
/home/vcap/app/venv/bin/pip install --upgrade pip-tools
/bin/bash: /home/vcap/app/venv/bin/pip: No such file or directory
make: *** [Makefile:37: freeze-requirements] Error 127
```

One approach we could take with this app (and Antivirus) would be to make this a `make freeze-requirements-in-docker` task like the others around it. That would address your concern about someone trying to run it locally - I agree it might be a surprise that everything else runs in Docker, but suddenly I need to have a local environment too.

Makefile
idavidmcdonald: Have rebased out

PR #591, opened 2021-11-18
Update Ghostscript & Poppler PDF libraries

rjbaker:
- Updates Ghostscript to 9.55.0
- Updates Poppler to 21.11.0

Signed-off-by: Richard Baker <richard.baker@digital.cabinet-office.gov.uk>


docker/Dockerfile
```diff
@@ -1,26 +1,19 @@
-FROM python:3.9-slim-buster as parent
+FROM python:3.9-slim-bullseye as parent
 
 ENV PYTHONUNBUFFERED=1
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN echo "Install base packages" && apt-get update \
     && apt-get install -y --no-install-recommends \
-        apt-transport-https \
```
rjbaker, line 8: This is no longer needed as it's included in apt by default these days: https://packages.debian.org/bullseye/apt-transport-https

benthorner (approved):
Nice cleanup! 🧹

While I'm nervous about the deployment based on the previous experience of others, objectively this PR looks good to me: the tests pass on the built container, and we have at least one test that covers [poppler](https://github.com/alphagov/notifications-template-preview/blob/6602e4557be7f2c55a46136c8f7352c8531eed49/tests/test_precompiled.py#L152) and one that covers [GhostScript](https://github.com/alphagov/notifications-template-preview/blob/6602e4557be7f2c55a46136c8f7352c8531eed49/tests/test_transformation.py).

We should check if [this bug](https://www.pivotaltracker.com/story/show/172077354) is fixed and potentially add a test to make sure we don't regress in future. That doesn't need to block this PR, but I've mentioned it in Slack to see if we can do it ASAP after this, so we don't regress in any future upgrade.

As for the fundamental changes:

- There are some dodgy-looking items in the GhostScript changelog, but I _think_ we'll be OK. We have lots of tests for the functionality we rely on. The part I'm nervous about is the other "random" changes piping entire PDFs through GS might make. We can't do anything about that, though, aside from getting sample PDFs to test (out-of-scope).

- Poppler is less clear, as we use it via `pdf2image`. I'm concerned about this one because [we don't actually have a test for the happy path](https://github.com/alphagov/notifications-template-preview/pull/584#discussion_r745544382). However, the impact of this is less severe than GhostScript: we may start rejecting or permitting PDFs when we shouldn't, but at least we don't have mangled them (that's my concern with GhostScript).

Short summary:

- Ideally we should add a couple more tests before this goes out, but it's not critical.
- We should definitely discuss a deployment / monitoring strategy before this goes out.

idavidmcdonald:
I haven't delved into all of the details but this looks good and I agree with Bens comment.

I think it is good that we aren't also trying to upgrade the python package dependencies in this PR, that can come as a separate change. 

In terms of deployment plan, I think we should look to release this change at about 4:30pm to production, not on a Friday.

I would be interested in discussing the idea of then rolling it back after 5:30pm to the previous template preview so we can then give DVLA some time to process and double check there are no issues. If after a few days we hear there are no problems we could then roll out for a whole day and then revert back again to check and then after that it should be all good to just stay out.

CrystalPea (approved):
Looks good to me, 🤞🏼  for no pdf shenanigans, but we have the recovery plan now, so we are that bit safer ☺️

benthorner:
Me and @rjbaker paired on testing this locally. We made [some temporary code changes](https://github.com/alphagov/notifications-template-preview/pull/new/temp-no-validation-test-visual-178758588) to:

- Disabling all the validations, so we could easily drive tests through the UI.
- Forcing all PDFs to pass through the two GhostScript commands we have.

Detailed results are below. **Short summary: this PR _will_ cause visual changes to PDFs we process, but these appear to be minor**. However, something that appears minor visually may have a big affect at print time e.g. if DVLA can't print them due to the new colour or font behaviour. However (however), there's realistically no more testing we can do without further investment in this app, so we will just have to proceed with caution.

_Note: it's possible some of the subtle font changes were just artefacts from rendering the PDF as an image. Even then, there is a definite decrease in the thickness of some fonts with this PR._

_Note 2: we also tested `landscape_rotated_page.pdf` but saw nothing remarkable before / after, beyond the same differences described below._

## Blacks are "black-er"

Left in both screenshots is the original PDF.

The effect is hard to see from the screenshots, although the "blue-r" one is clearly "brighter" in the centre.

| Before (blue-r)  | After (black-er) |
| old_black_comparison | new_black_comparison |

## Other colours after darker

| no_colour.pdf  | example_dwp_letter.pdf |
| https://user-images.githubusercontent.com/9029009/143279797-eef8c5aa-9e49-4b9c-97c4-bb22218276b2.mov | https://user-images.githubusercontent.com/9029009/143279856-da347981-0021-406f-8a48-7380b6c44007.mov

## Example font changes

| public_guardian_sample.pdf (changed)  | address_block_repeated_on_second_page.pdf (unchanged) |
| https://user-images.githubusercontent.com/9029009/143280272-15e2cc5d-9975-42c8-a83c-16c4c9797158.mov | https://user-images.githubusercontent.com/9029009/143280858-0c21038c-8edb-4f1b-81d7-348c957b32d5.mov |

benthorner:
Discussed further offline. After more testing we found the above differences are only in the preview PNGs and not the actual letter PDFs. We didn't check for this initially as in theory the PNG generation shouldn't have changed due to GhostScript or Poppler changes. However, on further investigation we found that the change to the base Docker image has had some impact on the versions of packages installed, including `libmagickwand-dev`, which accounts for the PNGs being different.

@rjbaker has updated the PR description with all the changed packages, including where they are used. The only additional one of concern is `libpango` / Weasyprint, which is used for generating templated letter PDFs from our HTML templates. We've visually checked that templated PDFs look the same before and after the changes, so no new concerns there.

[after_templated.pdf](https://github.com/alphagov/notifications-template-preview/files/7603800/after_templated.pdf)
[before_templated.pdf](https://github.com/alphagov/notifications-template-preview/files/7603801/before_templated.pdf)

Summary: nothing of concern, just superficial visual changes in PDF previews, ready to deploy.

PR #596, opened 2021-11-25
Add info logs to record major PDF transformations

benthorner:
https://www.pivotaltracker.com/story/show/178758588

This is more convenient than manually location, downloading and
processing the original PDF, and will help us understand what
proportion of PDFs are affected by these transformations.

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -221,14 +221,21 @@ def rewrite_pdf(file_data, *, page_count, allow_international_letters, filename)
         allow_international_letters=allow_international_letters,
     )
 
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):

benthorner:
Thanks for the review @idavidmcdonald.

Worth noting we don't actually have any tests that cover the conditional for unembedded fonts (proven by none of them triggering the new log output, which I had to test manually). The other new logs were triggered, but I think that's more incidental than intentional. I did start adding a test, but now isn't a great time so I'm backed out. Something to bring forward into the discussion of tech debt and this app, perhaps.

PR #598, opened 2021-11-29
Add guidance for deploying changes

benthorner:
https://www.pivotaltracker.com/story/show/178758588

This is based on our recent experience upgrading GhostScript and
Poppler [1] and may be useful for future changes.

👉 [Rendered version](https://github.com/alphagov/notifications-template-preview/blob/deploy-guidance-178758588/docs/deploying.md)

[1]: https://github.com/alphagov/notifications-template-preview/pull/591

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,4 @@ Celery is used for sanitising PDF letters asynchronously. It requires the `NOTIF
 ## Further documentation
 
 - [Making local requests](docs/local-requests.md)
+- [Guidance for deploying changes](docs/deploying.md)
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -1,0 +1,137 @@
+# Deploying
+
+Follow the guidance below when deploying changes that affect the:
+
+- [generation](#changes-to-pdf-generation)
+- [validation](#changes-to-pdf-validation) or
+- [preview](#changes-to-previewing-pdfs) of PDFs.
+
+The aim of the guidance is to help minimise risk for these kinds of changes.
+
+## Changes to PDF generation
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -107,3 +107,7 @@ Now follow the [instructions on the Wiki](https://github.com/alphagov/notificati
 ## To update application dependencies
 
 `requirements.txt` is generated from the `requirements.in` in order to pin versions of all nested dependencies. If `requirements.in` has been changed, run `make freeze-requirements` to regenerate it.
+
+## Further documentation
+
+- [Guidance for deploying changes](docs/deploying.md)
```
quis, line 113: I think it would make more sense if ‘Guidance for deploying changes’ was in under the ‘Deploying’ H2 in the README.

Might also be worth adding a one-liner about when you might want to do the extra work in the linked document. Presumably it’s not the intention that we do all that stuff for deploying any code change or dependency update?

docs/deploying.md
```diff
@@ -0,0 +1,133 @@
+# Deploying
+
+Follow the guidance below to deploy changes.
+
+Each section covers a different kind of change. The aim is to help minimise risk when deploying changes to this app, but you can always choose to skip some of the steps.
```
quis, line 5: It might be worth putting something here saying that it’s the PDF-generating/validating/reading bits of the code that we need to be a bit more careful around. I just wouldn’t want someone new to the team thinking that they need to do this testing to deploy any change. But I will leave this up to you.

quis (approved):

docs/deploying.md
benthorner: That's fair. My assumption was a new person would see there's no applicable section for their change, but that's a bit implicit. I've rewritten this to be more like a contents list.

What do you think?

quis (approved):

PR #604, opened 2021-12-14
Make code in deployment docs easier to reuse

benthorner:

This will make it faster to do future visual diff testing.

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -18,25 +18,41 @@ _*In a typical hour we see PDFs produced by e.g. PDFsharp, LibreOffice, PDFKit, 
 
 Test the output looks visually OK before / after with a few sample PDFs. You should consider [tweaking the code](https://github.com/alphagov/notifications-template-preview/commit/73b0b557429fad7a1280a691b583bf8d9c569d9c) so you can force any PDF through the full gamut of processing we may do to it.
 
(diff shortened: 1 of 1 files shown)


rjbaker (approved):

docs/deploying.md
```diff
@@ -18,25 +18,41 @@ _*In a typical hour we see PDFs produced by e.g. PDFsharp, LibreOffice, PDFKit,
 
 Test the output looks visually OK before / after with a few sample PDFs. You should consider [tweaking the code](https://github.com/alphagov/notifications-template-preview/commit/73b0b557429fad7a1280a691b583bf8d9c569d9c) so you can force any PDF through the full gamut of processing we may do to it.
 
-- [Remember to run `make bootstrap-with-docker`](https://github.com/alphagov/notifications-template-preview#docker-container) if switching between versions of dependencies. If you don't, the code will run with an out-of-date image, using old / random dependencies.
+```bash
+# reproduce changes without making a commit
+git cherry-pick -n 73b0b557429fad7a1280a691b583bf8d9c569d9c
```
leohemsted, line 23: i don't understand what this is doing - or more accurately why you would cherry pick a commit. wouldn't it be just checking out master, then checking out your branch?

benthorner: It's one more than that:

- I have a branch I want to test.
- I want to make import some temporary hacks to enable the testing.

But I don't want to actually add a commit to my branch with the test hacks. I also don't want to have to manually type out the test hacks myself / go and read what they are.

Does that help? 

I can clarify the comment, although it should be obvious what it does when it's run.

PR #605, opened 2021-12-15
Stop using Pillow (PIL) directly in our code

benthorner:

Please review commit by commit.

Every letter we send has a white / invisible "NOTIFY" tag in the 
top-left-hand corner. DVLA need this to mark the first page of 
each letter after they've been combined into a print run. They 
can't add it in for us, so we do it manually.

PDF coordinates start from the bottom-left-hand corner, so in 
order to position the tag, we need to know the height of a line 
of text. Historically, we also had the tag close to the left border 
(unclear why), and needed to know the text width to make sure 
it didn't actually cross our page border.

We used the Pillow (PIL) library to calculate the width / height 
of the "NOTIFY" tag text, so we could position it. Pillow is also 
required by a bunch of other dependencies e.g. reportlab. But 
just because it's installed doesn't mean we should use it:

- It's yet another library in a codebase of many libraries.
- [Our tests need to adapt when its behaviour changes](https://github.com/alphagov/notifications-template-preview/pull/572#issuecomment-976541325).

It turns out we can simplify how we calculate the position of 
the "NOTIFY" tag, so we don't need to use Pillow directly.

This PR makes five changes:

1. Redefines the tag bounding box to match what DVLA will accept.
2. Moves the tag further left so we no longer care about its width.
3. Manually calculates the height of the tag instead of using Pillow.
4. Moves the tag further up so it's consistent with the new "left".
5. Upgrades Pillow to the latest version, to resolve security issues.

Note that (2) and (3) are required to stop using Pillow directly, and 
(4) is just for completeness.

DVLA have confirmed the new position is OK from a sample, but 
we should still be cautious about deploying - especially with the 
upgrade to Pillow. See comments below for updates on that.

## Bounding box

unnamed

## Screenshots

Font colour temporarily changed to black.

| Before  | After
| Screenshot 2021-12-15 at 18 02 10 | Screenshot 2021-12-15 at 18 01 16 |


benthorner:
Testing with a precompiled letter from "Check your State Pension" ([seemingly the only service still injecting the tag themselves](https://github.com/alphagov/notifications-template-preview/pull/606#issuecomment-996646222)) shows the new bounding box still correctly identified the tag as present.

benthorner:
Validation testing complete [as per the guidance](https://github.com/alphagov/notifications-template-preview/blob/222071338ecd41754f58a3457b9888795de9ffcc/docs/deploying.md#changes-to-pdf-generation). I couldn't see any visual differences, apart from the expected shift in the NOTIFY tag. I plan to follow the full test deployment guidance:

- Other dependencies like `reportlab` depend on Pillow, so the upgrade could feasible have an effect on them.
- While DVLA have confirmed the change is OK, we don't change our PDFs often so we should still be cautious.

I'm not worried about the bounding box changes: the previous one was incorrect anyway.

leohemsted (approved):
think this looks good! deffo nicer to exactly mimick DVLA's check, nice one

benthorner:
All letters from the trial deployment are showing as delivered (except one, which got stuck at virus scanning).

Also checked hourly validation errors over the last few days, and there's nothing suspicious in the trial window.

Given the lack of indicators, I think it's worth the risk of just deploying this.

What do you think @leohemsted ?

leohemsted:
go for it i say 👍

PR #606, opened 2021-12-16
Add log about PDFs that contain "NOTIFY" tag

benthorner:

This is so we can see if it's OK to delete the code that handles
this scenario. Originally we asked services to inject the tag at
source, but now we do it for them. It's possible some PDFs still
have it in, so we shouldn't just remove the check.

I noticed the conditional wasn't actually covered by any tests,
so I've added one and a file that explicitly contains the tag.


benthorner:
Have verified this tag is still being injected by the folks at "Check your State Pension". There could be more, though, and it's hard to retain this kind of historical info, so I think it's still worth having the log.

benthorner:
Quick analysis of this indicates it's only the "Check your State Pension" service that's still doing this.

```
fields @timestamp, @message
| sort @timestamp desc
| filter message like "PDF already contains Notify tag"
| parse message "tag (NOTIFY.*." as ref
| display ref
| limit 1000
```

```
select distinct service_id from notifications where reference in (...);
```

PR #608, opened 2021-12-17
Simplify version control process for dodgy files

benthorner:

All the ignored files were PDFs, which we don't expect to change,
so the value in having them show up in a "git diff" is low. This
commit makes the process for changing a PDF the same as the process
for adding one - I think that's better since I forgot to follow
the current process when adding a new one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ app/version.py
 .envrc
 
 ## We don’t allow committing certain file types that risk containing users’ personal information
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
To be honest, I don't feel particularly strongly that this is an improvement over what we already have but I haven't recently worked with adding PDFs to template preview. Knowing that you have though, I'm happy to back your suggestion.

PR #611, opened 2021-12-21
Downgrade log for expected PDF exception

benthorner:
I've added a test to cover this. We do get other exceptions, one of
which I've included in the test:

    File "/home/vcap/app/app/precompiled.py", line 476, in _overlay_printable_areas_with_white
        page.mergePage(new_pdf.getPage(0))
      File "/usr/local/lib/python3.9/site-packages/PyPDF2/pdf.py", line 2239, in mergePage
        self._mergePage(page2)
      File "/usr/local/lib/python3.9/site-packages/PyPDF2/pdf.py", line 2248, in _mergePage
        originalResources = self["/Resources"].getObject()
      File "/usr/local/lib/python3.9/site-packages/PyPDF2/generic.py", line 516, in __getitem__
        return dict.__getitem__(self, key).getObject()
    KeyError: '/Resources'

While it's tempting to ignore these as well, it's unclear if they
represent an issue with the PDF or with the library. I've only seen
it happen once, so it doesn't seem worth the time to investigate at
the moment - the user still got a meaningful error.

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -12,6 +12,7 @@ from notifications_utils.postal_address import PostalAddress
 from pdf2image import convert_from_bytes
 from PIL import ImageFont
 from PyPDF2 import PdfFileReader, PdfFileWriter
+from PyPDF2.utils import PdfReadError
 from reportlab.lib.colors import Color, black, white
 from reportlab.lib.pagesizes import A4
 from reportlab.lib.units import mm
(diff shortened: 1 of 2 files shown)


app/precompiled.py
```diff
@@ -196,12 +196,10 @@ def sanitise_file_contents(encoded_string, *, allow_international_letters, filen
             "file": base64.b64encode(file_data.read()).decode('utf-8')
         }
     except Exception as error:
-        if isinstance(error, ValidationFailed):
```
CrystalPea, line 199: Nice one doing the work to reduce non-actionable errors. Wouldn't it be better though to add known errors to list of errors that only cause a warning, and if the error falls outside of those, still cause an exception? Then we would see if something unexpected happens.

benthorner: I thought about this some more and agree it's worth keeping the error-level log, but for a specific reason: if we suddenly have a large volume of unexpected exceptions, we will get notified via Elastalert.

I also thought more about `KeyError` and don't think it's best to keep erroring for that one: it probably is a bug with our code or the library we're calling, and ideally we should invest some time in trying to fix it.

I've rewritten the PR based on those thoughts.

What do you think?

CrystalPea (approved):

CrystalPea:
I think that's a good approach after the changes - we will get less unhelpful errors, but we are still prepared for the unexpected.

PR #614, opened 2021-12-30
Trial running Sentry in a Celery app

benthorner:
https://www.pivotaltracker.com/story/show/180766893

See https://github.com/alphagov/notifications-admin/pull/4114.

### DIFF
--- a/run_celery.py
+++ b/run_celery.py
@@ -1,7 +1,20 @@
 #!/usr/bin/env python
 
-from app import notify_celery, create_app  # noqa
+import os
 
+import sentry_sdk
+from sentry_sdk.integrations.celery import CeleryIntegration
+
+from app import create_app, notify_celery  # noqa
+
+if 'SENTRY_DSN' in os.environ:
+    sentry_sdk.init(
+        dsn=os.environ['SENTRY_DSN'],
+        integrations=[CeleryIntegration()],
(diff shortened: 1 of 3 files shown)


run_celery.py
```diff
@@ -1,7 +1,19 @@
...
+
+from app import create_app, notify_celery  # noqa
+
+sentry_sdk.init(
+    dsn=os.environ['SENTRY_DSN'],
+    integrations=[CeleryIntegration()],
+    environment=os.environ['NOTIFY_ENVIRONMENT'],
+    attach_stacktrace=True,
+    traces_sample_rate=0.00005  # avoid exceeding rate limits in Production
+)
```
idavidmcdonald, line 16: Any reason that we don't have the equivalent `sentry_sdk.set_level('error') # only record error logs or exceptions` that we have in the admin version of this PR?

benthorner: Good catch! Turns out that was a mistake on my part. I've also put this whole block in a conditional to support local development. Here's a PR for Admin with some more explanation: https://github.com/alphagov/notifications-admin/pull/4121

leohemsted (approved):

run_celery.py
benthorner: I'm going to merge now, but let me know if you want to discuss more David.

PR #616, opened 2022-01-05
Fix log format to avoid double interpolation

benthorner:
This was trying to log the set of unembedded fonts, which has the
same form as a format string - e.g. {'/Helvetica'} - and was being
picked up by an old, magical bit of handler code in utils [1].

Although the log gets through, this stops the excess of error logs
that don't make it to Kibana, but do make it to Sentry.

[1]: https://github.com/alphagov/notifications-utils/blob/aa3b918e6c6c6d8b94b5aaf3c47d69680e7f1611/notifications_utils/logging.py#L159

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -243,7 +243,7 @@ def rewrite_pdf(file_data, *, page_count, allow_international_letters, filename)
         file_data = convert_pdf_to_cmyk(file_data)
 
     if unembedded := contains_unembedded_fonts(file_data, filename):
-        current_app.logger.info(f'PDF contains unembedded fonts: {unembedded}')
+        current_app.logger.info(f'PDF contains unembedded fonts: {", ".join(unembedded)}')
         file_data = embed_fonts(file_data)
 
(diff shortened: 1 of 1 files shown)


leohemsted (commented):
this is fine, but i suspect if you want to look at this you'll be playing whackamole across the codebase and i'd rather fix the logger in utils than try and have unwritten rule "never put curly brackets in log statements"

can you just dismiss the exception in sentry for now?

benthorner:
> can you just dismiss the exception in sentry for now?

I can, but it will still contribute towards our rate limit. At this rate we'll surpass our free tier limit in days, instead of the month it's supposed to last for. I wasn't expecting this, based on the logs that actually make it to Kibana.

This is actually the only such issue - at least, according to Sentry. The log is also optional - we don't use it for anything - so I'd even prefer to remove it if we're not keen on fixing it.

I do appreciate it's only a superficial fix, though.

leohemsted (approved):
approved on the basis that it's killing our free allowance while we're just playing around with sentry. but my old comment still stands long term I think

PR #603, opened 2021-12-14
Bump reportlab from 3.5.67 to 3.6.3

benthorner:
https://www.pivotaltracker.com/story/show/180421592

Changelog: https://fossies.org/linux/reportlab/CHANGES.md

All changes look minor, no cause for concern.

PR #617, opened 2022-01-05
Update address block boundaries to match spec

benthorner:
Relates to email thread: "Alignment Issue - Hackney"

DVLA insert a long string - the "MDI" - just above the recipient
address, so it's important we reject letters with content that's
straying above this area. The example DVLA gave has quite a lot of
overlap - the entire first line of the address.

Two reasons for the overlap:

- The boundary box we're using is out-of-date. This updates the box
to match the current spec [1].

- We allow 1mm extra all around the address box [2], apparently to
avoid pixel-level validation issues.

I think it's the combination of these that's causing the problem,
but I don't want to open a can of worms by looking at the second
one. Hopefully correcting the raw bounding box will be enough to
start rejecting letters we're currently accepting.

[1]: https://drive.google.com/file/d/1vR4-02TlzA-VzXPhSrTr3MtUP5wog4x4/view
[2]: https://github.com/alphagov/notifications-template-preview/blob/1c3cbd13f5b4fa5caa10bb6ae30644f574c880aa/app/precompiled.py#L434


quis:
I’m a bit worried that this is a breaking change for users of our API. In [our specification](https://docs.notifications.service.gov.uk/documentation/images/notify-pdf-letter-spec-v2.4.pdf) we tell people that the top and bottom edges of the address area are 39.5mm and 66.3mm respectively:

image

---

Because of the 1mm buffer the this PR changes the boundary we check from 38.5mm to 39mm. So anyone falling foul of our check would still be falling foul of our published specification.

On the other hand, someone may have an integration that is working now and not overlapping the MDI. If we break the integration and they don’t have an easy way to make changes on their end then we could be disrupting important letters.

quis:
Update: we shouldn’t need to do this because we rewrite the address to put it in the proper position.

If DVLA are seeing an address in the wrong place it’s because it’s failed to be rewritten somehow.

See also https://github.com/alphagov/notifications-template-preview/pull/357

benthorner:
Thanks for your thoughts @quis ⭐. 

Just a quick reply to say I've checked this locally to confirm it's not due to https://github.com/alphagov/notifications-template-preview/pull/603, which was deployed just before this letter was submitted. The issue predates the upgrade.

benthorner:
I've had a look into this and it turns out we **don't** always rewrite the address block. Specifically, when we get a `RedactionException`, which bubbles up into a `redaction_failed_message` in `sanitise_file_contents`:

```
        return {
            "recipient_address": recipient_address,
            "page_count": page_count,
            "message": None,
            "invalid_pages": None,
            "redaction_failed_message": redaction_failed_message,
            "file": base64.b64encode(file_data.read()).decode('utf-8')
        }
```

What happens to this? It gets received in two places:

1. [In Admin, which ignores it entirely](https://github.com/alphagov/notifications-admin/blob/53164837f9e289e068779ccc4adf11a6d7ded577/app/main/views/uploads.py#L182-L224).
2. [In Template Preview Celery, which still lets the letter through](https://github.com/alphagov/notifications-template-preview/blob/68be67402bc9f891d681edf7e9a075faec3cdbf5/app/celery/tasks.py#L37-L40).

I'm struggling to see why we do this from the code history. I recall talking to @klssmith about whether we actually use the copies we make in (2), but not about why we let them through. @leohemsted do you know?

@quis I agree this PR is redundant, it's ☝️ that's the problem. That's nice we don't need to worry about breaking things for users. I think it's worth adding a comment around those constants to indicate they're special: unlike the other constants, they don't need to conform to DVLA's spec because we rewrite that part of the page.

I've made a separate PR for that, so we can continue discussing ☝️ here: https://github.com/alphagov/notifications-template-preview/pull/620

leohemsted:
What we do is select all text in the address area.
We then delete the original text.
Then we slap a white box over the top just in case there's any non-text artifacts remaining, and re-write that original text in arial font size 8 aligned properly etc etc.

However, there a variety of ways this might fail. This happens regularly but infrequently (554 times in the last week). Mostly it fails because deleting the original text is really hard and we don't want to accidentally alter content elsewhere in the file, so we err on the side of caution.

Rather than rejecting the letter with a 400 error, we decided to just send through the letter without touching the address area as a lesser-of-two-evils decision. We made this call, partly since this was an additional sanitisation we added on top of our PDF letter flow when it was already live. (We didn't want to start failing letters that we previously accepted). It's really hard to explain this stuff to services - especially when we're often not sure why these letters are failing. These aren't services with deep and intimate knowledge of postscript and how to craft a letter - a lot of them are just letters that their existing case working system spits out that they have no control over the creation of, they just upload it to Notify.

We store the original file in case we want to debug why a given letter failed to redact. I suspect we looked at this a lot more when we were actively and frequently working on the redaction code.

---------------------

I'm open to us doing something clever like:

```python
do_margin_validation()
try:
  rewrite_address_block()
except RedactionException:
  # make sure that the MDI specifically won't be overwritten
  do_margin_validation_with_accurate_address_boundary()
```

That way, most letters will continue as normal. If they're slightly outside the boundaries of the address box, normally it doesn't matter as we fix the address box ourselves. However, if we can't fix the address box we can try again just to make sure it won't cover the MDI.

benthorner:
Thanks for the context @leohemsted ⭐. 

For reference, I found these two commit messages helpful:

- [Transition to using a redactor](https://github.com/alphagov/notifications-template-preview/commit/df52416ef05ddaac607065d5627cb53a42675f10#diff-7f3ab9275b78a2f680054fd0a80f7952bca8b09151ee0c9f02035a5f0bec3048).
- [How we got to the current state](https://github.com/alphagov/notifications-template-preview/commit/28fd34e16d72ecc0cd5d344ec90084b5ad89b49f#diff-7f3ab9275b78a2f680054fd0a80f7952bca8b09151ee0c9f02035a5f0bec3048)

That last commit also hints at this being the least worst option, like you said. I'm minded to write some documentation for this, as it's the overall process is complex and subtle.

However, I'm also wondering if we can just do a better job of redacting - ideally without the (somewhat) vendored code we have now. I had a go at this in https://github.com/alphagov/notifications-template-preview/pull/621 and the results look promising.

Some options:

1. Treat #621 as documentation and link to it in the README. Leave it at that for now. I'd like to clarify to DVLA that we won't actually be making any changes to this.

2. Write a card to look at #621 properly and, if it actually looks viable, turn it into a proper PR and carefully try deploying it.

3. Contact the Hackney service to see if they can provide a sample PDF that also fails redaction. We could use it for an xfail test to document the current state of the world.

I think I'll just do (3) as it's a quick win. What do you think we should do next?

benthorner:
Closing this as I think it's served it's purpose:

- I'm in contact with the Hackney service and will look at a separate PR to add a test around the issue here.
- We can use the draft PR above for further discussion about this - it's at least worth a bug card.

PR #621, opened 2022-01-07
Try using PyMuPDF ("fitz") library for redaction

benthorner:
Previously we used a vendored / customised PDF redactor for this
purpose, but it doesn't always work [1]. In this circumstance, we
rely on DVLA being able to manually print these letters [2], which
isn't ideal. Using the PyMuPDF redactor should always work, as it
uses a simpler technique of wiping all text in the box [3], while
the previous redactor tried to find the first address-like string.

This is just a test commit - to make testing easy, I've hooked up
the "/preview" endpoint so it runs a full sanitise on page #1.

A possible flaw with this approach is that it will wipe any other
text that it will wipe any other content in the address box, but
that _should_ be fine because there can't be any: we're using the
same box we use to extract the address [5], so we should never get
as far as redacting the address if there is any extraneous text in
there, as the letter would be marked as invalid because of it.

Still, this is something that would need more scrutiny. In addition,
we're now passing each PDF through the PyMUPDF library, which may
make other dodgy changes - I haven't seen anything bad in my tests.

## Screenshots

The green fill is just to show where the redaction is happening. I 
manually tested with the PDF mentioned in https://github.com/alphagov/notifications-template-preview/pull/617 :

- Confirmed the old redactor fails to redact it.
- Confirmed redaction works with PyMUPDF.

| Old redactor | PyMUPDF |
| - | - |
| Screenshot 2022-01-07 at 15 36 39 | Screenshot 2022-01-07 at 15 36 08 |

## Things to do

- [x] Decide if we want to pursue this work (bug card / impact / effort)
- [ ] Update DVLA (email subject: "Alignment Issue - Hackney")

[1]: https://github.com/alphagov/notifications-template-preview/pull/617#issuecomment-1006760959
[2]: https://github.com/alphagov/notifications-template-preview/commit/28fd34e16d72ecc0cd5d344ec90084b5ad89b49f#diff-7f3ab9275b78a2f680054fd0a80f7952bca8b09151ee0c9f02035a5f0bec3048
[3]: https://pymupdf.readthedocs.io/en/latest/page.html#Page.add_redact_annot
[4]: https://pymupdf.readthedocs.io/en/latest/page.html#Page.apply_redactions
[5]: https://github.com/alphagov/notifications-template-preview/blob/8946aec2bf890af25fe7d869d36e4c34deb1d03a/app/precompiled.py#L702-L711

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -768,21 +768,32 @@ def _get_pages_with_notify_tag(src_pdf_bytes):
 
 
 def redact_precompiled_letter_address_block(pdf, address_regex):
-    options = pdf_redactor.RedactorOptions()
+    # options = pdf_redactor.RedactorOptions()
 
-    options.content_filters = []
-    options.content_filters.append((
-        re.compile(address_regex),
-        lambda m: " "
-    ))
-    options.input_stream = get_first_page_of_pdf(pdf)
-    options.output_stream = BytesIO()
+    # options.content_filters = []
+    # options.content_filters.append((
+        # re.compile(address_regex),
+        # lambda m: " "
+    # ))
+    # options.input_stream = get_first_page_of_pdf(pdf)
+    # options.output_stream = BytesIO()
 
-    pdf_redactor.redactor(options)
+    # pdf_redactor.redactor(options)
 
-    options.output_stream.seek(0)
+    # options.output_stream.seek(0)
 
-    return replace_first_page_of_pdf_with_new_content(pdf, options.output_stream)
+    # return replace_first_page_of_pdf_with_new_content(pdf, options.output_stream)
+    pdf.seek(0)
+    doc = fitz.open("pdf", pdf)
+    page = doc[0]
+    x1 = ADDRESS_LEFT_FROM_LEFT_OF_PAGE - 3
+    y1 = ADDRESS_TOP_FROM_TOP_OF_PAGE - 3
+    x2 = ADDRESS_RIGHT_FROM_LEFT_OF_PAGE + 3
+    y2 = ADDRESS_BOTTOM_FROM_TOP_OF_PAGE + 3
+    rect = fitz.Rect(x1*mm, y1*mm, x2*mm, y2*mm)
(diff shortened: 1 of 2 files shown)


CrystalPea (commented):
Wow, if this works and really fully redacts the text, not just covering it, that would be amazing!

When I was implementing the current PDF redaction code, redaction by area was what I was looking for, but couldn't find a tool erasing the text fully, just tools that covered the text with white rectangle, which was what we were already doing.

benthorner:
Thanks @CrystalPea. Yes, it's an exciting development 😃.

I was also a bit sceptical, but I've done an experiment to reassure myself:

- I reused [part of the deployment instructions](https://github.com/alphagov/notifications-template-preview/blob/master/docs/deploying.md#manual-testing) to process a PDF before and after the change and save the results to separate files - one file being how we do it now, and one with the change in this PR.
- I also commented out [this line](https://github.com/alphagov/notifications-template-preview/blob/222071338ecd41754f58a3457b9888795de9ffcc/app/precompiled.py#L619) and ran the same testing code to produce a PDF with one address hidden beneath the other (like you mentioned in your comment).
- I than ran all three PDFs through the following code (adapted from extracting the address block). For the before/after PDFs there is only one instance of "nceyshktvrai", whereas for the overlap PDF there are two.

```
import fitz
f = open('over_hackney_sample.pdf', 'rb')
doc = fitz.open('pdf', f.read())
words = doc[0].get_text_words()
[w[4] for w in words if 'nceyshktvrai' in w[4]] # the fourth element is the actual text
```

It could be the original address is still hidden in the PDF somewhere, but there's no evidence of this. To get a little more reassurance, I've sent two test PDFs to DVLA to see if they can print them:

- One PDF was processed using the new approach in this PR (ideally this prints OK).
- One PDF was not redacted and the original address is hidden (ideally this confirms previous issues).

CrystalPea:
Hey @benthorner what you did to test this new way is really thorough and methodical, and looks really promising, well done! Fingers crossed that results from DVLA are as expected, that would mean our process can be improved <3

PR #622, opened 2022-01-10
Fix image build due to broken GhostScript version

benthorner:
To fix this error:

      21.53 The following packages have unmet dependencies:
      21.67  ghostscript : Depends: libgs9 (= 9.53.3~dfsg-7+deb11u1) but 9.53.3~dfsg-7+deb11u2 is to be installed
      21.69 E: Unable to correct problems, you have held broken packages.

Changelog: https://www.mail-archive.com/debian-changes@lists.debian.org/msg23791.html

I think this is happening because deb11u1 refers to the specific OS
version, which has changed in the base image (although I can't work
out how to get the full OS version to confirm this).

Bumping the package version will fix the issue for now, although we
should consider locking the base image to a specific tag, if we want
to avoid this happening again. Arguably we should continue as we are
until we have a security upgrade policy for our images.

Given this is a security update, I'm not concerned about the deploy
and don't think we need to do any further testing.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,7 +12,7 @@ RUN echo "Install base packages" && apt-get update \
         libpango1.0-dev \
         libmagickwand-dev \
         imagemagick \
-        ghostscript=9.53.3~dfsg-7+deb11u1 \
+        ghostscript=9.53.3~dfsg-7+deb11u2 \
         poppler-utils=20.09.0-3.1 \
         xfonts-utils \
         gsfonts \


rjbaker (approved):
Nice one. Yes looks like it's just a security dependency update. Pinning ghostscript in this way ensures we know when things have changed, rather than being silently upgraded between builds.

Pinning the base image would be good, but it's essential we have properly configured Dependabot (etc) to ensure we receive security updates to the base image. Waiting for Notify to make a decision on that.

benthorner:
Thanks @rjbaker. Agree this is something we could have a process for, so I've added it to [the discussion doc we have for how we manage dependencies](https://docs.google.com/document/d/1zUwLYBMrMV0YdhKwY3KwXMqAJ3Hq9hkeOrugHHq1eok/edit), which we should be revisiting soon.

PR #624, opened 2022-01-12
Add test to document redactor bug

benthorner:
Ideally redaction should work for all PDFs where we can extract the
address block, hence adding this test using xfail.

Due to the complexity of the redactor code, I'm going to leave fixing
the bug to follow-up work [1]. The failing test is a net improvement
by itself: there are lots of unknowns about the various packages we
use to rewrite PDFs, so the more tests the better!

[1]: https://github.com/alphagov/notifications-template-preview/pull/621


CrystalPea (approved):
Nice one 👍🏼

PR #625, opened 2022-01-13
Log filename when letter redaction fails

benthorner:
https://www.pivotaltracker.com/story/show/180900091

This will allow us to investigate the scale of the issue. We see
thousands of these logs every week, but:

- It's unclear how many services are affected.
- We can't investigate individual failures.

Having the letter filename will allow us to gather more data to fix
the bug, which may actually be several bugs.


CrystalPea (approved):
that will be very useful, good work 👍🏼

PR #637, opened 2022-02-25
Downgrade error logs for all 400 responses

benthorner:
This tidies up the existing error handlers and stops us
getting error logs for unactionable scenarios.

Please review commit by commit.


app/__init__.py
```diff
@@ -203,6 +203,11 @@ def exception(error):
             # error.code is set for our exception types.
             return jsonify(result='error', message=str(error)), 500
 
+    @app.errorhandler(400)
+    def bad_request(error):
+        app.logger.warning(str(error))
```
idavidmcdonald, line 208: So in theory, I'm OK with this being a warning level log.

However, in practice, I think we should keep this as an error level log. This is because the only people calling template preview should be us, and if we are getting 400s then it means something is not working as expected. A single one of these should cause us to investigate/log a bug about it. If we downgrade this to a warning, there is little chance we'd see it and little chance we'd get reminded of the fact there is a bug (or if there were similar bugs introduced in the future).

There could be an argument to downgrade this to a warning and instead raise the error at the client end calling template preview instead (I assume that isn't already happening).

Note, `error` is appropriate rather than `exception` as `exception` will give you the stack trace too which I don't think is relevant for us in this case.

Also note, I may be missing or misunderstanding something here with this particular change. I think the intention is to stop us getting emails each time we get a 4xx on template preview but do correct me where I'm wrong!

app/precompiled.py
```diff
@@ -517,7 +517,10 @@ def _colour_no_print_areas_of_single_page_pdf_in_red(src_pdf, is_first_page):
     :param BytesIO src_pdf: A file-like representing a single page pdf
     :param bool is_first_page: true if we should overlay the address block red area too.
     """
-    pdf = PdfFileReader(src_pdf)
+    try:
+        pdf = PdfFileReader(src_pdf)
+    except PdfReadError as e:
```
idavidmcdonald, line 522: I can see that the `binascii.Error` will no longer be caught by a specific handler and instead be handled by the default exception handler.

However, I can see that we will be catching `PdfReadError`s here and then raising an `InvalidRequest`.

Why do we have different behaviour for the two different exceptions? Why not catch either of them here? Or leave both of them to go uncaught?

idavidmcdonald (commented):
Happy with the middle commit but two questions please to help me have a better understanding behind the changes to make sure I'm on the right page!

app/__init__.py
benthorner: Fair point. I assumed the missing handler for `400` was an omission, but perhaps the `Exception` handler is meant to pick it up. I think having an explicit handler is clearer though.

I've added a commit to fix this here: https://github.com/alphagov/notifications-template-preview/pull/637/commits/ae8dcde0e2f450456b7e83b0001a377d25cf6673

## So why are we getting this error?

I did some more investigation and found a way to reproduce it:

1. Create a template with placeholders that (just) wraps onto a second page.
2. Prepare to send a letter but use short values so the letter is just one page.
3. Instead of sending, go "back" a couple of steps (to the address field).
4. See that you get a "400 Bad Request: Letter does not have a page 2" error.

### Screenshots

| Step 1 | Step 2 | Step 3 |
| - | - | - |
| screencapture-notifications-service-gov-uk-services-3b525922-bdf9-4589-9d7b-dfe969f57805-send-7084e849-c427-41af-8942-f6224befed30-one-off-address-2022-03-01-11_57_13 |  screencapture-notifications-service-gov-uk-services-3b525922-bdf9-4589-9d7b-dfe969f57805-send-7084e849-c427-41af-8942-f6224befed30-one-off-step-7-2022-03-01-11_56_45 | screencapture-notifications-service-gov-uk-services-3b525922-bdf9-4589-9d7b-dfe969f57805-template-7084e849-c427-41af-8942-f6224befed30-notification-check-2022-03-01-11_57_00 |

## What to do next

This PR could still go in (amended as above).

I can write this up as a low impact/effort bug card.

How does that sound?

app/precompiled.py
idavidmcdonald: Just to check I understood, you wanted to leave both of them uncaught but you found an existing test for the `PdfReadError` and decided that provided enough reason for you to explicitly catch it, rather than say deleting the test?

And then also you say you wanted to leave both of them uncaught but in the sentence above you said that in general we should avoid global exception handlers unless the exception is global and these aren't. That seems to say to me to be contradictory (you say you want them uncaught but before it said that these are not good candidates for the global exception handler) or maybe I've misunderstood your comment...

app/__init__.py
idavidmcdonald: Yes please on writing a bug card. Great work on debugging this, I don't think anyone has been able to work this one for ages what was causing it.

I agree on the low impact. I haven't given it enough thought to know how much effort it would be to fix but I'm happy to go by your suggestion of low effort too.

idavidmcdonald (approved):
I think this is likely good. I'm not 100% sure on it and its repercussions but I'm sure any issues would be low impact enough if I have misunderstood something.

I think your fourth commit is a contender to be fixed up in to the third commit with a rewritten commit message for better git history as the commit before goes in one direction and then this takes it a bit back. I don't mind if you don't do this though.

app/__init__.py
idavidmcdonald: Also thanks for taking on board the log level stuff I raised initially!

app/precompiled.py
benthorner: > you say you want them uncaught but before it said that these are not good candidates for the global exception handler

Ah, I should have been clearer when I said "uncaught".

I mean that they should be caught by the general global exception handler, not the specific global exception handlers I'm removing in the first commit. It's really another instance of [your comment about the 400 handler](https://github.com/alphagov/notifications-template-preview/pull/637#discussion_r815916397): these handlers have been downgrading all  `PdfReadError` and `binascii.Error` exceptions to a 400 response / warning log, but sometimes the exception may actually represent a bug and we'd miss it.

> decided that provided enough reason for you to explicitly catch it, rather than say deleting the test?

The test is evidence that we need to handle the exception at a specific place in the code, in addition to any other places where we handle it already. But there's no evidence that these exceptions are always benign. The fact that we don't seem to get any of them in Production means there's little risk of removing the global handlers that currently downgrade the logs for all such exceptions - irrespective of how they occurred.

Is that making more sense now?

benthorner:
> I think your fourth commit is a contender to be fixed up in to the third commit

Good point. I've taken on board [your comment about the strategy here](https://github.com/alphagov/notifications-template-preview/pull/637#discussion_r821910460) and rewritten the commit message to explain it.

app/precompiled.py
idavidmcdonald: I think it is making a bit more sense. I don't know if I understand it well enough to reexplain it to someone else but I think that is OK.

Thank you for taking the time on this.

PR #645, opened 2022-03-28
Change log level from error to exception

idavidmcdonald:
These exceptions are unexpected and it is important to see the
strack trace when debugging however using `error` hides the
stack trace. Moving it to to `exception` means we include the
stack trace.

This will be helpful as at the moment I am seeing log lines of

```
Unexpected exception for precompiled pdf: IndexError('tuple index out of range') for file name: 70a42a07-7089-45b5-a135-7f6fd77ec673
```

however there is no stacktrace to give me context.


benthorner (approved):
Good call 👍

PR #647, opened 2022-04-07
Bump non pdf dependencies

klssmith:
This bumps all the test dependencies, the production dependencies that aren't related to PDFs, and removes an unused dependency - defusedxml.


idavidmcdonald (approved):
Split nicely in to small commits, thanks

PR #650, opened 2022-04-13
Prepare to switch to Redis on PaaS

benthorner:
https://www.pivotaltracker.com/story/show/181796569

This adds similar code to [^1]. Since this app didn't
previously load CloudFoundry config - and the config
code was also very different to other apps - there are
some initial commits to lay a foundation first.

The config module is still different to other repos due 
to the heavy use of NOTIFY_ENVIRONMENT to avoid 
duplication. I think it would be too subjective to change 
the approach in this PR - leaving as-is for now.

[^1]: https://github.com/alphagov/notifications-admin/pull/4210

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,11 +1,8 @@
-import json
-import os
 from contextlib import suppress
 from hashlib import sha1
 
 from flask import Flask, jsonify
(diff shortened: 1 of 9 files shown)


benthorner:
Closing due to a major oversight: this app doesn't use Redis 😝

PR #651, opened 2022-04-13
Standardise config into a app/config.py module

benthorner:
For some reason this app is different to all our other
repos in the way it loads and applies config. This makes
a bunch of tweaks to get it looking similar.

The config module is still different to other repos due
to the heavy use of NOTIFY_ENVIRONMENT to avoid
duplication. I think it would be too subjective to change
the approach in this PR - leaving as-is for now.

Note: this refactor came from work to use PaaS Redis [^1]
before I reaised this app doesn't use Redis!

[^1]: https://github.com/alphagov/notifications-template-preview/pull/650

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -16,14 +16,15 @@ class TaskNames:
 
 
 class Config:
+    NOTIFY_ENVIRONMENT = os.environ['NOTIFY_ENVIRONMENT']
+
     AWS_REGION = 'eu-west-1'
     TEMPLATE_PREVIEW_INTERNAL_SECRETS = json.loads(
         os.environ.get('TEMPLATE_PREVIEW_INTERNAL_SECRETS', '[]')
     )
-    NOTIFY_ENVIRONMENT = os.environ['NOTIFY_ENVIRONMENT']
     NOTIFY_APP_NAME = 'template-preview'
-    DANGEROUS_SALT = os.environ['DANGEROUS_SALT']
-    SECRET_KEY = os.environ['SECRET_KEY']
+    DANGEROUS_SALT = os.environ.get('DANGEROUS_SALT')
+    SECRET_KEY = os.environ.get('SECRET_KEY')
 
     CELERY = {
         'broker_url': 'sqs://',
@@ -51,13 +52,13 @@ class Config:
     # it will still create the config key with None value causing
     # logging initialization in utils to fail
     if 'NOTIFY_LOG_PATH' in os.environ:
-        NOTIFY_LOG_PATH = os.environ['NOTIFY_LOG_PATH']
+        NOTIFY_LOG_PATH = os.environ.get('NOTIFY_LOG_PATH')
 
     EXPIRE_CACHE_IN_SECONDS = 600
 
-    if os.environ['STATSD_ENABLED'] == "1":
+    if os.environ.get('STATSD_ENABLED') == "1":
         STATSD_ENABLED = True
-        STATSD_HOST = os.environ['STATSD_HOST']
+        STATSD_HOST = os.environ.get('STATSD_HOST')
         STATSD_PORT = 8125
     else:
         STATSD_ENABLED = False
--- a/app/config.py
+++ b/app/config.py
@@ -50,12 +50,9 @@ class Config:
 
     EXPIRE_CACHE_IN_SECONDS = 600
 
-    if os.environ.get('STATSD_ENABLED') == "1":
-        STATSD_ENABLED = True
-        STATSD_HOST = os.environ.get('STATSD_HOST')
-        STATSD_PORT = 8125
-    else:
-        STATSD_ENABLED = False
+    STATSD_ENABLED = True
+    STATSD_HOST = os.environ.get('STATSD_HOST')
+    STATSD_PORT = 8125
 
 
 class Production(Config):
@@ -102,6 +99,8 @@ class Preview(Config):
 
 class Development(Config):
     NOTIFY_ENVIRONMENT = 'development'
+
+    STATSD_ENABLED = False
 
     LETTERS_SCAN_BUCKET_NAME = 'development-letters-scan'
     LETTER_CACHE_BUCKET_NAME = 'development-template-preview-cache'
--- a/app/config.py
+++ b/app/config.py
@@ -24,19 +24,15 @@ class Config:
     DANGEROUS_SALT = os.environ.get('DANGEROUS_SALT')
     SECRET_KEY = os.environ.get('SECRET_KEY')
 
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -78,3 +79,16 @@ def set_config(app, name, value):
     app.config[name] = value
     yield
     app.config[name] = old_val
+
+
+@pytest.fixture
+def os_environ():
+    # for use whenever you expect code to edit environment variables
+    old_env = os.environ.copy()
+    os.environ.clear()
```
leohemsted, line 88: why are we clearing here before the yield? this leads to surprising behaviours (like having to redefine the pytest.ini provided variables like NOTIFY_ENVIRONMENT) and isn't consistent with the api fixture.

tests/test_config.py
```diff
@@ -0,0 +1,23 @@
+import importlib
+import os
+
+import pytest
+
+from app import config
+
+
+@pytest.fixture
+def reload_config(os_environ):
```
leohemsted, line 10: hmmm i'd be tempted to just bundle the `os_environ` fixture in to this. i'm not convinced we'll need to use os_environ much outside of testing the config loading itself, given we already have patterns for setting config values themselves temporarily for tests and that seems like it'll have much less potential side-effects and interesting quirks with needing to reload imports etc.

app/config.py
```diff
@@ -0,0 +1,80 @@
...
+
+    CELERY = {
+        'broker_url': 'sqs://',
+        'broker_transport_options': {
+            'region': AWS_REGION,
+            'visibility_timeout': 310,
+            'wait_time_seconds': 20,  # enable long polling, with a wait time of 20 seconds
+            'queue_name_prefix': ({
+                'test': 'test',
+                'development': os.environ.get('NOTIFICATION_QUEUE_PREFIX', 'development'),
```
leohemsted, line 37: i don't think we should default to `'development'` as that'll leave people potentially using that variable and overwriting each other. Ideally we should error but that might not be easy to write since this is all executed at module level on every env 🤔. In other apps we just let it default to `None`.

app/config.py
```diff
@@ -0,0 +1,80 @@
...
+        NOTIFY_LOG_PATH = os.environ.get('NOTIFY_LOG_PATH')
+
+    EXPIRE_CACHE_IN_SECONDS = 600
+
+    if os.environ.get('STATSD_ENABLED') == "1":
+        STATSD_ENABLED = True
+        STATSD_HOST = os.environ.get('STATSD_HOST')
+        STATSD_PORT = 8125
+    else:
+        STATSD_ENABLED = False
```
leohemsted, line 64: it might be nice to replicate the existing other app configs here and just have this without the if statement if that works

```
    STATSD_HOST = os.getenv('STATSD_HOST')
    STATSD_PORT = 8125
    STATSD_ENABLED = bool(STATSD_HOST)
```

leohemsted (commented):
Seems sensible. A few things I think we could clean up a bit more though - all of them are pretty optional so if you disagree or don't think it's worth it that's fine!

I think I would prefer our config to be separated like on the other apps, even if it means that the bucket names are a bit more duplicative and prescriptive - it's very useful to be able to ctrl+f to find things like bucket names and domains. It also avoids excessive logic ran at module level at import time, which I think is a generally decent aim.

app/config.py
```diff
@@ -0,0 +1,80 @@
...
+        'task_queues': [
+            Queue(QueueNames.SANITISE_LETTERS, Exchange('default'), routing_key=QueueNames.SANITISE_LETTERS)
+        ],
+    }
+
+    # if we use .get() for cases that it is not setup
+    # it will still create the config key with None value causing
+    # logging initialization in utils to fail
+    if 'NOTIFY_LOG_PATH' in os.environ:
+        NOTIFY_LOG_PATH = os.environ.get('NOTIFY_LOG_PATH')
```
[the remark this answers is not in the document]
benthorner: I tried this but it _is_ actually required (to my surprise as well):

```
app/__init__.py:39: in create_app
    utils_logging.init_app(application, application.statsd_client)
/usr/local/lib/python3.9/site-packages/notifications_utils/logging.py:28: in init_app
    ensure_log_path_exists(app.config['NOTIFY_LOG_PATH'])
/usr/local/lib/python3.9/site-packages/notifications_utils/logging.py:46: in ensure_log_path_exists
    Path(path).parent.mkdir(mode=755, parents=True)
/usr/local/lib/python3.9/pathlib.py:1082: in __new__
    self = cls._from_parts(args, init=False)
/usr/local/lib/python3.9/pathlib.py:707: in _from_parts
    drv, root, parts = self._parse_args(args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'pathlib.PosixPath'>, args = (None,)

    @classmethod
    def _parse_args(cls, args):
        # This is useful when you don't want to create an instance, just
        # canonicalize some constructor arguments.
        parts = []
        for a in args:
            if isinstance(a, PurePath):
                parts += a._parts
            else:
>               a = os.fspath(a)
E               TypeError: expected str, bytes or os.PathLike object, not NoneType

/usr/local/lib/python3.9/pathlib.py:691: TypeError

It's tempting but I don't think it's worth the effort to investigate, so leaving as-is.

Note: I specifically used `os.getenv` to match Antivirus, so I expected it to behave the same.

tests/test_config.py
benthorner: Fair enough, I remember thinking the same. 

I think it's helpful to keep it separate to tease out the two things it does.

Changed in https://github.com/alphagov/notifications-template-preview/pull/651/commits/4fa5c3d0b0b3a7e36b52a137129144fc8ce03757.

app/config.py
benthorner: Sure, changed in https://github.com/alphagov/notifications-template-preview/pull/651/commits/d754c2d3c7300a09efa5d41386f6c82748a6a6dd.

app/config.py
benthorner: Sure, changed in https://github.com/alphagov/notifications-template-preview/pull/651/commits/d3a5ab9d87c87d9df7ce4b17e5d1be26f25c9634.

tests/test_config.py
leohemsted: Nice compromise 👍

app/config.py
leohemsted: ahhhhh the live production mixup strikes again!

leohemsted (approved):
👍

PR #652, opened 2022-04-13
Fix branch deployments from M1 Macs to PaaS

benthorner:
An image built on a Mac M1 machine won't run on a PaaS AMD instance,
so we need to use Docker's new buildx feature [^1].

Note: you may need to run "docker buildx create --use" first, but the
command will prompt you to do so if necessary.

[^1]: https://docs.docker.com/buildx/working-with-buildx/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,11 +71,10 @@ test-with-docker: ## Run tests in Docker container
 
 .PHONY: upload-to-docker-registry
 upload-to-docker-registry: ## Upload the current version of the docker image to Docker registry
-	docker build -f docker/Dockerfile -t ${DOCKER_IMAGE_NAME} .
 	$(if ${DOCKER_USER_NAME},,$(error Must specify DOCKER_USER_NAME))
 	$(if ${CF_DOCKER_PASSWORD},,$(error Must specify CF_DOCKER_PASSWORD))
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
Looks reasonable.

I'm currently running NOT on an M1 mac but one of the older (Intel?) macs.

I've tested this by running `docker buildx build --platform linux/amd64 -f docker/Dockerfile -t david-test .` Note, I didn't test the `push` side of things but assume that is fine. Things built successfully. I assume that is good enough but let me know if there is any further due diligence you think I should do.

Question that I'm sure the answer to won't change my approval. There are two changes in here.

1. use `buildx build` instead of `build`
2. move from the build and the push being does as separate steps into a single step

For step 2, is this just a refactoring you did as you went to tidy things up or is there a more specific reason behind this?

benthorner:
Thanks for taking a look David!

I don't think there's anything more you need to test, as I've successfully done a branch deploy using this new recipe.

> For step 2, is this just a refactoring you did as you went to tidy things up or is there a more specific reason behind this?

This is necessary as the built image isn't actually made available to your system - if you're using buildx then it's likely it won't work on your machine anyway. This means using the old `docker push` technique won't work - it all has to be done with `buildx`. (I do also like that it's a bit more succinct than before.)

PR #661, opened 2022-05-16
Run integration tests on multiple CPU cores

quis:
As we do in other apps<sup>1 2</sup> we can make the tests run faster by executing them across multiple worker processes in parallel.

`-n auto` tells `pytest-xdist` to spawn the same number of worker processes as the machine has CPUs.

***

Before | After
---|---
image | image

***

1. https://github.com/alphagov/notifications-admin/blob/e467167c62ac4ba5d599e26570b1e01290f2ca1b/Makefile#L64
2. https://github.com/alphagov/notifications-api/blob/318cc1284aefa6f8e30928c1a0f703d5e5cc3a59/Makefile#L72

### DIFF
--- a/tests/test_precompiled.py
+++ b/tests/test_precompiled.py
@@ -209,47 +209,48 @@ def test_get_invalid_pages_second_page(x, y, expected_failed, client):
         assert get_invalid_pages_with_message(packet) == ('', [])
 
 
-@pytest.mark.parametrize('x, y, page, expected_message', [
-    (0, 0, 1, ('content-outside-printable-area', [1])),
-    (200, 200, 1, ('', [])),
-    (590, 830, 1, ('content-outside-printable-area', [1])),
-    (0, 200, 1, ('content-outside-printable-area', [1])),
-    (0, 830, 1, ('content-outside-printable-area', [1])),
-    (200, 0, 1, ('content-outside-printable-area', [1])),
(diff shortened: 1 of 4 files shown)


tests/test_precompiled.py
```diff
@@ -209,47 +209,48 @@ def test_get_invalid_pages_second_page(x, y, expected_failed, client):
...
-
-    cv.setStrokeColor(black)
-    cv.setFillColor(black)
-    cv.setFont('Arial', 6)
-    cv.drawString(x, y, 'This is a test string used to detect non white on a page')
-
-    cv.save()
-    packet.seek(0)
-    assert get_invalid_pages_with_message(packet) == expected_message
+def test_get_invalid_pages_black_text(client):
```
klssmith, line 212: Does changing the test like this make it run faster with pytest-xdist?

klssmith (approved):

tests/test_precompiled.py
quis: This line raised an exception with `parametrize` – presumably the underlying code isn’t thread safe or something like that:
https://github.com/alphagov/notifications-template-preview/pull/661/files#diff-19b9f7e801b6d6bd10f6ec60266969a3ac9e281555b8afb688864f94e89b695fR247-R248

PR #699, opened 2022-10-24
Bump PyPDF2 to 2.10.9 to fix KeyError

samuelhwilliams:
We've been seeing quite a few KeyError `/Resources` errors lately. This can happen when pages don't have a `/Resources` key - but in PDFs missing resources can be inferred from previous pages. This issue was raised https://github.com/py-pdf/PyPDF2/pull/1349 and fixed in v2.10.9.

Bumping to this version will stop the app throwing 500s and instead either complete successfully or return the correct validation error, which should prevent some users contacting support.


samuelhwilliams:
concourse isn't building the test image correctly (with the new dependency 🤔)

tests/test_precompiled.py
```diff
@@ -763,3 +763,31 @@ def test_redact_precompiled_letter_address_block_only_touches_first_page():
...
+    client, auth_header
+):
+    """
+    This tests to make sure that PyPDF2 doesn't raise a KeyError when sanitising a PDF that is missing /Resources
+    on one of the pages. Resources should be inferrer from one of the parent/previous pages in this case.
+
+    The PDF under test was provided by one of our services when they encountered the error. Ideally the PDF would
+    be valid and return a successful response, but some test that this returns a correct response is at least better
+    than this call raising an error/500."""
+    with open("tests/test_pdfs/no_resources_on_last_page.pdf", "rb") as pdf:
```
klssmith, line 778: You can use the the approach we have for other test PDFs in `pdf_consts.py`, which provides a helper function for opening the file and also means that all test PDFs get documented there.

klssmith (approved):

CrystalPea (approved):
Nice one 👏🏼 . We have been seeing this error for so long in our logs, great that it's being fixed/handled now 🥳

PR #719, opened 2023-03-30
Validate and sanitise letter attachment

CrystalPea:
When we validate and sanitise an attachment, we don't have to do certain things:
- we don't have a first page with different boundaries, so all pages have same boundary validation
- we don't have an address block to rewrite
- we don't add any tags

Story ticket: https://trello.com/c/qJ82a3L9/292-validate-and-sanitise-letter-attachment-uploaded-by-user
Admin PR: https://github.com/alphagov/notifications-admin/pull/4664

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -253,6 +253,19 @@ def rewrite_pdf(file_data, *, page_count, allow_international_letters, filename)
         filename=filename,
     )
 
+    file_data = normalise_fonts_and_colours(file_data, filename)
+
+    # during switchover, DWP and CYSP will still be sending the notify tag. Only add it if it's not already there
+    if not is_notify_tag_present(file_data):
+        current_app.logger.info("PDF does not contain Notify tag, adding one.")
+        file_data = add_notify_tag_to_letter(file_data)
+    else:
+        current_app.logger.info(f"PDF already contains Notify tag ({filename}).")
+
+    return file_data, recipient_address
+
+
+def normalise_fonts_and_colours(file_data, filename):
     if not does_pdf_contain_cmyk(file_data):
         current_app.logger.info("PDF does not contain CMYK data, converting to CMYK.")
         file_data = convert_pdf_to_cmyk(file_data)
@@ -267,14 +280,7 @@ def rewrite_pdf(file_data, *, page_count, allow_international_letters, filename)
         )
         file_data = embed_fonts(file_data)
 
-    # during switchover, DWP and CYSP will still be sending the notify tag. Only add it if it's not already there
-    if not is_notify_tag_present(file_data):
-        current_app.logger.info("PDF does not contain Notify tag, adding one.")
-        file_data = add_notify_tag_to_letter(file_data)
-    else:
-        current_app.logger.info(f"PDF already contains Notify tag ({filename}).")
-
-    return file_data, recipient_address
+    return file_data
 
 
 @precompiled_blueprint.route("/precompiled/overlay.png", methods=["POST"])
@@ -470,6 +476,33 @@ def _overlay_printable_areas_with_white(src_pdf):
     """
 
     pdf = PdfReader(src_pdf)
+
+    _overlay_printable_areas_of_address_block_page_with_white(pdf)
+
+    # For each subsequent page its just the body of text
+    for page_num in range(1, len(pdf.pages)):
+        page = pdf.pages[page_num]
+
+        can = NotifyCanvas(white)
+
+        # Each page of content
+        pt1 = BORDER_LEFT_FROM_LEFT_OF_PAGE - 1, BORDER_TOP_FROM_TOP_OF_PAGE - 1
+        pt2 = BORDER_RIGHT_FROM_LEFT_OF_PAGE + 1, BORDER_BOTTOM_FROM_TOP_OF_PAGE + 1
+        can.rect(pt1, pt2)
+
+        # move to the beginning of the StringIO buffer
+        new_pdf = PdfReader(can.get_bytes())
+
+        page.merge_page(new_pdf.pages[0])
+
+    out = bytesio_from_pdf(pdf)
+    # it's a good habit to put things back exactly the way we found them
+    src_pdf.seek(0)
+
+    return out
+
+
+def _overlay_printable_areas_of_address_block_page_with_white(pdf):
     page = pdf.pages[0]
     can = NotifyCanvas(white)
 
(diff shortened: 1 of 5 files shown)


tests/test_precompiled.py
```diff
@@ -487,11 +506,18 @@ def test_precompiled_sanitise_pdf_with_notify_tag(client, auth_header):
     assert is_notify_tag_present(pdf)
 
 
+@pytest.mark.parametrize(
+    "is_an_attachment",
```
samuelhwilliams, line 510: optional nit: is `query_string` a better name?

app/precompiled.py
```diff
@@ -196,16 +197,22 @@ def sanitise_file_contents(encoded_string, *, allow_international_letters, filen
...
-            file_data,
-            page_count=page_count,
-            allow_international_letters=allow_international_letters,
-            filename=filename,
-        )
+        if is_an_attachment:
+            file_data = normalise_fonts_and_colours(file_data, filename)
+            recipient_address = None
+        else:
+            file_data, recipient_address = rewrite_pdf(
```
samuelhwilliams, line 210: Did you consider passing `is_an_attachment` into this function and decide not to do it? It feels like we are having to pull out some of that function's internals into this function now - which is maybe a bit less clean?

I can imagine `rewrite_pdf` getting a bit messy if we pass that flag in - could we maybe add a `rewrite_attachment_pdf` with the same interface and just the operations we want? It would keep both of those functions implementations simple, and keep their internal decisions about what rewriting needs to take place out of this function.

samuelhwilliams (commented):
Nice work on the main logic, which all looks good. Also appreciate the test coverage 🙌 

I like a lot some of the refactoring in principle although I think there's a couple of bits we could go further/differently. Most of the comments are just around code style and clarity, so hopefully just minor changes :)

app/precompiled.py
CrystalPea: Yeah, I considered passing `is_an_attachment` to `rewrite_pdf`, but I decided it would be clunky because of the order of operations (where attachments need only the middle two operations).

I also quite like how it got divided: `normalise_fonts_and_colours` deals with general preparation of the PDF for printing, while `rewrite_pdf` does operations related to more DVLA-specific requirements.

CrystalPea: If I rename `normalise_fonts_and_colours` to `rewrite_subsequent_pages`, would that work better?

app/precompiled.py
```diff
@@ -726,22 +751,25 @@ def is_notify_tag_present(pdf):
...
+def _get_pages_with_notify_tag(src_pdf_bytes, is_an_attachment=False):
     """
     Looks at all pages except for page 1, and returns any pages that have the NOTIFY tag in the top left. DVLA can't
     process letters with NOTIFY tags on later pages because their software thinks it's a marker signifying when a new
     letter starts. We've seen services attach pages from previous letters sent via notify
     """
     src_pdf_bytes.seek(0)
     doc = fitz.open("pdf", src_pdf_bytes)
-    if doc.pageCount == 1:
+    ignore_first_page = 1
```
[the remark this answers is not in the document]
CrystalPea: hmm could rename it to `starting_page_index`?

app/precompiled.py
```diff
@@ -174,13 +172,16 @@ def sanitise_precompiled_letter():
...
+        is_an_attachment=request.args.get("is_an_attachment"),
     )
     status_code = 400 if sanitise_json.get("message") else 200
 
     return jsonify(sanitise_json), status_code
 
 
-def sanitise_file_contents(encoded_string, *, allow_international_letters, filename):
+def sanitise_file_contents(
+    encoded_string, *, allow_international_letters, filename, is_an_attachment=False
```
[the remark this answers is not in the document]
CrystalPea: Sorry but I don't understand, what would be the benefit of specifying type here and making sure that the value is actually True/False, not just truthy/falsey? Like, aren't non-bool objects being truthy/falsey so we can use them like this?

app/precompiled.py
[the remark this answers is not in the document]
CrystalPea: Renamed ☺️

app/precompiled.py
```diff
@@ -470,6 +487,36 @@ def _overlay_printable_areas_with_white(src_pdf):
     """
```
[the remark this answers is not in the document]
CrystalPea: ahh this is where it makes a difference if the object is true Boolean or just truthy/falsey then? Hard to say in the docstring what it is at present 😅

app/precompiled.py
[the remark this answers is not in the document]
samuelhwilliams: strings being truthy/falsey can be nice, but it's also inherently less precise. My issue here isn't strictly about truthy/falsey-ness, it's about the fact that the function signature heavily implies that `is_an_attachment` is a bool value (both in the name and the default value), but in reality it's basically never _actually_ a bool. It just adds extra mental load and is a potential gotcha when reading through the code.

If we have a thing that represents a yes/no value, it makes the most sense for it to be the data structure that represents a yes/no, ie a bool.

I'm aware that we use truthy/falseyness elsewhere, but the above compounded by the fact that we're passing `is_an_attachment=true` in URL params again makes it feel like we should be able to pass `is_an_attachment=false` - but if we do that, then the `false` string value is 'truthy'. Which is a bit unintuitive and another potential gotcha.

app/precompiled.py
samuelhwilliams: Hmmm ... I'm not sure about that, because we'd be calling `rewrite_subsequent_pages` on _all_ of the PDF's pages if the document is an attachment.

CrystalPea: but the attachment on the whole is subsequent pages of a letter.

CrystalPea: Still, I quite like as it is now, so happy to leave it ☺️

samuelhwilliams: Yes, I realise. We're just debating semantics really. I'm just letting you know I still find it confusing - you can rethink it or not :)

app/precompiled.py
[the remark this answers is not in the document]
CrystalPea: Ahh, thanks a lot for explaining!

app/precompiled.py
```diff
@@ -174,13 +172,16 @@ def sanitise_precompiled_letter():
         encoded_string,
         allow_international_letters=allow_international_letters,
         filename=request.args.get("upload_id"),
+        is_an_attachment=request.args.get("is_an_attachment"),
```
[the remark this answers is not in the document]
CrystalPea: I did the same thing we do for `allow_international_letters` value, where the value defaults to False if the arg we get is anything else than `"true"`.

samuelhwilliams (approved):

PR #727, opened 2023-05-02
Connect to sqs using the new endpoint

klssmith:
Use the new SQS endpoints which will allow us to use VPC endpoints in ECS.

This matches what we do in https://github.com/alphagov/notifications-api/pull/3787


sakisv (approved):
nice 👍

PR #728, opened 2023-05-02
Use stable secure version of imagemagick

klssmith:
When building the Docker image we were getting errors that `Version '8:6.9.11.60+dfsg-1.3' for 'imagemagick' was not found`. Version `8:6.9.11.60+dfsg-1.3+deb11u1` has been created to fix a security vulnerabiltiy, so this changes the version in the Dockerfile.


sakisv (approved):
Nice finding 👍

PR #737, opened 2023-06-07
Create/preview templated letter pdfs with attachment

leohemsted:
If a letter attachment json blob is passed through when either previewing via `POST /preview.pdf` or creating via the `create_pdf_for_templated_letter` task, append that PDF to the generated letter PDF. Note that the two generated letter PDFs are different when previewing and creating (since the preview has fake QR codes and the print version has the white NOTIFY text).

This code needs the API to send over the new fields before we can merge it.

- [x] https://github.com/alphagov/notifications-api/pull/3818

Once we've merged this PR, we then need to update admin/api when they call `/preview.pdf` to pass through the `letter_attachment` json blob. We already pass it through when you download a pdf from the check page before sending[^1], but not when you download a pdf of a created notification[^2]

[^1]: https://github.com/alphagov/notifications-admin/blob/main/app/main/views/send.py#L740
[^2]: https://github.com/alphagov/notifications-admin/blob/main/app/main/views/notifications.py#L215

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -11,6 +11,7 @@ from notifications_utils.template import LetterPrintTemplate
 
 from app import notify_celery
 from app.config import QueueNames, TaskNames
+from app.letter_attachments import add_attachment_to_letter
 from app.precompiled import sanitise_file_contents
 from app.preview import get_page_count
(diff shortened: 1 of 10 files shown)


samuelhwilliams (approved):
This looks good to me and is working locally as far as I see 👍

PR #739, opened 2023-06-28
Adding an endpoint for previewing a page for a letter attachment

whpearson:
This adds an endpoint to to preview one page for a letter attachment, without associating it with a template (as it may just be uploaded)

There are tests for out of range requests and bad auth, as well as the happy path.

### DIFF
--- a/app/preview.py
+++ b/app/preview.py
@@ -13,7 +13,7 @@ from wand.image import Image
 
 from app import auth
 from app.letter_attachments import add_attachment_to_letter, get_attachment_pdf
-from app.schemas import get_and_validate_json_from_request, preview_schema
+from app.schemas import get_and_validate_json_from_request, letter_attachment_preview_schema, preview_schema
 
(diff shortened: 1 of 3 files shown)


app/preview.py
```diff
@@ -138,6 +138,41 @@ def view_letter_template(filetype):
...
+
+    if requested_page <= attachment_page_count:
+        encoded_string = base64.b64encode(attachment_pdf)
+        png_preview = get_png_from_precompiled(
+            encoded_string=encoded_string,
+            page_number=requested_page,
+            hide_notify=False,
+        )
+    else:
+        abort(400, f"Letter does not have a page {requested_page}")
```
samuelhwilliams, line 168:
```suggestion
        abort(400, f"Letter attachment does not have a page {requested_page}")
```

samuelhwilliams (approved):
Tested the endpoint locally and looks good 👍 

Just a typo tweak would be nice

PR #741, opened 2023-06-29
Bump PyMuPDF version and replace deprecated pageCount property

BlessedDev:
PyMuPDF now makes wheels for MacOS available on pypi.org and automatically
 downloads the required MuPDF source and builds it into PyMuPDF.
This solves previous installation problems where the `setup.py` file could not find specific terms (`build,bdist`, `bdist`_`wheel`) in `sys.argv` to initiate
 downloading and building MuPDF locally.
 The `pageCount` property name is now deprecated and the new version is `page_count`.

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -729,7 +729,7 @@ def _get_pages_with_notify_tag(src_pdf_bytes, is_an_attachment=False):
     starting_page_index = 1
     if is_an_attachment:
         starting_page_index = 0
-    if doc.pageCount == starting_page_index:
+    if doc.page_count == starting_page_index:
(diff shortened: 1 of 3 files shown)


CrystalPea (approved):

CrystalPea (changes requested):
Actually, I think requirements.in need to be changed and then make freeze-requirements run instead of changing requirements.txt directly.

BlessedDev:
> Actually, I think requirements.in need to be changed and then make freeze-requirements run instead of changing requirements.txt directly.

Good catch. I have done that now.

CrystalPea (approved):

PR #744, opened 2023-07-20
Update weasyprint to render QR codes in letter

CrystalPea:
Weasyprint only started SVGs since version 53, and at first it was a bit buggy, so I updated to the latest version.

See this issue: https://github.com/Kozea/WeasyPrint/issues/75

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -15,7 +15,7 @@ pypdf2==2.10.9
 reportlab==3.6.3
 pdf2image==1.12.1
 PyMuPDF==1.22.5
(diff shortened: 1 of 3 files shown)


samuelhwilliams (approved):
This looks good 👍 

We checked to understand why the cache keys have changed and if it was concerning. It's not.

Some of the HTML is syntatically different - for example `margin: 16pt 0 0 0;` vs `margin: 16.0pt 0 0 0;`, but functionally and visually this is the same.

PR #748, opened 2023-07-25
Bump utils to pull in pypdf change

samuelhwilliams:
And gunicorn, since we're bumping that on a bunch of other apps too.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -86,7 +86,7 @@ geojson==2.5.0
     # via notifications-utils
 govuk-bank-holidays==0.11
     # via notifications-utils
-gunicorn==20.1.0
+gunicorn==21.2.0
     # via -r requirements.in
 html5lib==1.1
     # via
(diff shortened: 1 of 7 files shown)


leohemsted (approved):

leohemsted (approved):

requirements.txt
```diff
@@ -143,10 +145,10 @@ pydyf==0.7.0
     # via weasyprint
 pymupdf==1.22.5
     # via -r requirements.in
+pypdf==3.13.0
+    # via notifications-utils
 pypdf2==2.10.9
```
leohemsted, line 150: is this going to cause big issues? should we migrate this to use pypdf [3]?

leohemsted: especially since we did the migration to solve a security vuln warning right?

samuelhwilliams: yes 🙌  I rushed 😢

leohemsted (approved):

PR #752, opened 2023-08-08
Refactor the docker image

samuelhwilliams:
An initial dockerfile update. If/when this goes in, I'd like to make sure we're following a similar/consistent approach in all of our other dockerfiles.

---

Headlines:
* Production image size from 1030MB -> 780MB.
* Fewer dependencies installed and available in production image; smaller attack surface.
* App no longer running as `root` user
* venv is readonly in production image (read-write in test/dev image)

We do a number of things here:

A `notify` user
---------------

Create and switch to a `notify` user, rather than running the app as `root` user. There are various obvious security improvements that flow out of not using the root user.

Additional build stages
-----------------------

Add some extra stages to the build process. We split out the `parent` step into separate `base` and `python_build` stages.

The `base` step takes care of everything that needs to be run as the `root` user and are about installing OS-level dependencies, fonts, config etc. (**note**: if we go ahead with renaming this step we'll need to update the `base_image_build_for_docker_app` ci pipeline. Maybe it's not worth changing the name here?)

The `python_build` step handles creating the virtual environment, which means we can just copy the `venv` directory to a different clean stage. There can be space improvements here if things required to build/install the dependencies aren't actually needed post-build. This approach is recommended as best practice by various parties, including Snyk, a security-tooling company
(https://snyk.io/blog/best-practices-containerizing-python-docker/). There are a number of OS-level dependencies that we are only installing for the python requirements build/install, so by separating those out into a different stage we can exclude them from our production image.

Use a venv
----------

Finally, we install python dependencies into a virtual environment rather than to the default directory (/usr/local/bin). We make this venv readonly for the production image, but readwrite by the `notify` user for the test image (which will be used for local dev).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -49,15 +49,18 @@ USER notify
 
 RUN mkdir /home/vcap/logs
 
-COPY --from=python_build --chown=notify:notify /opt/venv /opt/venv
+COPY --from=python_build --chown=root:root /opt/venv /opt/venv
 ENV PATH="/opt/venv/bin:${PATH}"
 
 COPY --chown=notify:notify entrypoint.sh .
 ENTRYPOINT [ "/home/vcap/app/entrypoint.sh" ]
 
 ##### Test Image ##############################################################
-
 FROM ${BASE_IMAGE} as test
+
+USER root
+RUN chown -R notify:notify /opt/venv
+USER notify
 
 # Make sure the app/ directory is there so that "make bootstrap" can create app/version.py
 RUN mkdir -p app
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 ARG BASE_IMAGE=ghcr.io/alphagov/notify/notifications-template-preview:base
-FROM python:3.9-slim-bullseye as parent
+FROM python:3.9-slim-bullseye as base
 
 ENV PYTHONUNBUFFERED=1
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -26,21 +26,31 @@ RUN echo "Install base packages" && apt-get update \
...
-
-COPY policy.xml entrypoint.sh ./
+COPY policy.xml .
 
 # Overwrite the default ImageMagick policy which doesn't allow reading or writing PDFs
 RUN rm /etc/ImageMagick-6/policy.xml && cp ./policy.xml /etc/ImageMagick-6/policy.xml
+WORKDIR /home/vcap/app
+
+# Build/Install python dependencies
+FROM base AS python_build
```
sakisv: Why do we want to use the `python_build` step here? Is there any particular benefit to put it in a separate step instead of expanding the `base` step to:

1. Install all system packages
2. Create notify and celery* users
3. Install requirements as the notify user

That way we avoid copying and `chown`ing things

*Why do we need a celery user along with a notify user? wouldn't it be simpler to only have a single low-privileges user?

sakisv (commented):
I agree on standarising how we build our Docker images and this is definitely a step in the right direction 👍 

Just one question about the high level approach and the different steps in the Dockerfile

docker/Dockerfile
samuelhwilliams: > *Why do we need a celery user along with a notify user? wouldn't it be simpler to only have a single low-privileges user?

Good question - we probably don't? I just didn't think to remove it .. for some reason. Incase `celeryuser` was expected somewhere else I guess... I will remove it 👍

sakisv: Oh I hadn't realised that we were using it already 🤔 
Still, probably better to simplify anyway 😄

samuelhwilliams: Yeah, I guess that was the workaround at the time to get rid of an aggressive celery warning that is thrown if you try to run celery as root 😉 

---

Re the `python_build` step - the benefits might've been more obvious if I actually made good use of it 😂  The idea is that a number of the OS-level things we install are only required for installing/building the various python packages we need. If we build them in a separate stage, and then just copy the result, we are doing something similar to the `frontend_build` steps elsewhere and removing a load of packages from the prod build that aren't needed there. This results in a smaller attack surface (no dev tooling installed+available) and a smaller image (about 970MB -> 770MB) with change 1c013d32c93eaf69b27e50395e684cb9df6c51d0.

samuelhwilliams:
Sorry @sakisv. Made another change following on from your comments (and my tweaks) on `python_build`. Have just removed the `app_base` stage altogether and now have the `test` step derive from the `production` step. I think this is a little bit cleaner - one less stage to keep in mind at least. And feels like by deriving from the prod image, the tests that run are likely to be more representative of the actual prod container.

471845665310542ab634a07a998f52cb4f3ffb28

sakisv (approved):
Nice, thanks for explaining (and amending!) the use of various steps, it does make sense now 👍

PR #753, opened 2023-08-09
Add back curl/deps for celery

samuelhwilliams:
Otherwise we get errors:

   2023-08-09T12:44:26.98+0100 [APP/PROC/WEB/0] ERR raise ImportError('The curl client requires the pycurl library.')
   2023-08-09T12:44:26.98+0100 [APP/PROC/WEB/0] ERR ImportError: The curl client requires the pycurl library.
   2023-08-09T12:44:29.26+0100 [APP/PROC/WEB/0] ERR scripts/run_app_paas.sh: line 45: kill: (62) - No such process

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,6 +8,7 @@ RUN echo "Install binary app dependencies" \
     && apt-get update && \
     apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
+        libcurl4 \
         libpango1.0-dev=1.46.2-3 \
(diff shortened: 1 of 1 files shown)


sakisv (approved):
Would just `libcurl` be enough or we actually use curl too? 🤔

samuelhwilliams:
It's a good question - will try it out 🤞

samuelhwilliams:
@sakisv good call - I built and deployed dcda8842c26c96e9f95306c3b9e2120dccd5ead3 on preview and it does pass the template-preview functional tests, so that seems a better option 👍

PR #754, opened 2023-08-09
Generate version file and copy it

samuelhwilliams:
Let's generate the app version file in the `python-build` step and copy it across, so that we don't need to temporarily copy the `.git` dir in the prod build stage, or have access to `git`.


sakisv (approved):
ha nice and obvious (in hindsight) improvement 👍

PR #759, opened 2023-08-22
Double memory for template preview to give it more CPU

idavidmcdonald:
In production, the celery worker in particular, is spiking over 100% CPU very regularly. This is leading to the app crashing regularly.

There is no direct control for how much cpu to give a PaaS app but instead you increase the memory and this increases the available CPU.

https://grafana-notify.cloudapps.digital/d/_GlGBNbmk/notify-apps?orgId=1&var-space=production&var-app=notify-template-preview-celery&from=now-24h&to=now&refresh=1m&viewPanel=8

image

### DIFF
--- a/manifest-celery.yml.j2
+++ b/manifest-celery.yml.j2
@@ -3,7 +3,7 @@ applications:
 - name: notify-template-preview-celery
 
   disk_quota: 2G
-  memory: 2G
+  memory: 4G
   health-check-type: process
   command: exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=4 2> /dev/null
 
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
Should we include a comment (if we can?) to say that memory is this high for CPU reasons rather than actual memory reasons. Just so we don't think it's way too high and lower it again at some point in the future?

sakisv (approved):

idavidmcdonald:
Yeah, I think we can get away without a comment

- this code will go away soon
- I don't think I've ever seen us reduce memory for an app and we won't be focussing on cost saving measures on the PaaS at all
- hopefully people will be using git commit history to help guide them on what to take into consideration on values like these

PR #760, opened 2023-08-23
Chown /opt/venv faster

samuelhwilliams:
Copying the directory via a Dockerfile command is faster than running a command inside the build container to update ownership


whpearson (approved):
👍

PR #774, opened 2023-11-03
add `choom-wrap-pkg`, use an OOM adjustment of 50 for imagemagick, ghostscript, poppler binaries

risicle:
When the oom-killer kicks in, it would be preferable for it to target the "leaf" processes that the python app is executing to service a task. These deaths should be comparatively recoverable, whereas if the oom-killer targets one of our python processes it is likely to bring the whole instance down, producing collateral damage to any other tasks that were being processed on the instance and potentially losing logs.

So rather than tracking down every place that the app (or, more likely, one of the libraries it uses) invokes an os binary, wrap those binaries in a script that will adjust the oom-score for every invocation.

I've only done cursory tests of this so far. Also I have no idea if 50 is a good number - picked it as a relatively conservative starter.

### DIFF
--- a/docker/choom-wrap-pkgs.sh
+++ b/docker/choom-wrap-pkgs.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+# usage: choom-wrap-pkgs <adjustment> <pkg_name> ...
+# wraps binaries belonging to listed debian packages
+# in a choom(1) call making adjustment <adjustment>
+# to the process's oom score when executed.
+
+function choom_wrap_bin() {
+  local adjustment="$1"; shift
+  local bin_path="$1"; shift
+  echo "choom-wrapping ${bin_path} with adjustment ${adjustment}" >&2
+
+  local bin_basename="$(basename ${bin_path})"
+  local bin_dirname="$(dirname ${bin_path})"
+  local bin_newpath="${bin_dirname}/.${bin_basename}-wrapped"
+
+  mv "${bin_path}" "${bin_newpath}"
+
+  cat > "${bin_path}" <<EOF
+#!/usr/bin/env bash
+exec choom -n '${adjustment}' -- '${bin_newpath}' "\$@"
+EOF
+  chmod --reference="${bin_newpath}" "${bin_path}"
+}
+
+function choom_wrap_deb_pkg() {
+  local adjustment="$1"; shift
+  local pkg_name="$1"; shift
+  echo "choom-wrapping binaries from debian package ${pkg_name}" >&2
+
+  dpkg -L "${pkg_name}" | while read -r filepath ; do
+    if [ -f "${filepath}" -a -x "${filepath}" ] ; then
+      choom_wrap_bin "${adjustment}" "${filepath}"
+    fi
+  done
+}
+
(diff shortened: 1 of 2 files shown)


docker/choom-wrap-pkgs.sh
```diff
@@ -0,0 +1,47 @@
...
+
+  local bin_basename="$(basename ${bin_path})"
+  local bin_dirname="$(dirname ${bin_path})"
+  local bin_newpath="${bin_dirname}/.${bin_basename}-wrapped"
+
+  mv "${bin_path}" "${bin_newpath}"
+
+  cat > "${bin_path}" <<EOF
+#!/usr/bin/env bash
+exec choom -n '${adjustment}' -- '${bin_newpath}' "\$@"
```
samuelhwilliams, line 21:
```
notify@9e37384254ba:/home/vcap/app$ sleep 500
^Z
[1]+  Stopped                 sleep 500
notify@9e37384254ba:/home/vcap/app$ choom -p $(pgrep sleep)
pid 1144's current OOM score: 666
pid 1144's current OOM score adjust value: 0
notify@9e37384254ba:/home/vcap/app$ fg
sleep 500
^C
notify@9e37384254ba:/home/vcap/app$
notify@9e37384254ba:/home/vcap/app$ choom -n 50 -- sleep 500
^Z
[1]+  Stopped                 choom -n 50 -- sleep 500
notify@9e37384254ba:/home/vcap/app$ choom -p $(pgrep sleep)
pid 1147's current OOM score: 700
pid 1147's current OOM score adjust value: 50
```

Will this _increase_ the choom score and lead it to be targeted for death more easily, rather than decrease it? Should we be doing something like `choom -n -50` instead? Or am I misreading/misinterpreting?

samuelhwilliams: Ok yes i was misunderstanding. It _will_ increase the OOM score, but as you said, for the child processes like `gs`, where killing those is preferrable to killing the python process. Gotcha.

risicle: I should reiterate I have no idea whether this will _actually_ have a useful effect.

samuelhwilliams: Yeah. I've been mulling over whether we should instead _decrease_ the OOM score for the python process(es), but I also am not confident of confirming a good solution or not. 'The internet' also seems to be mixed on the topic, some suggesting that this way lies pain, it's more art than science, etc. Some seem to even suggest disabling oomkiller altogether. I'm not sure. Maybe it's worth a shot., but I'm not sure how we'd convince ourselves that it's doing what we expect 🤔

risicle: The reason I didn't decrease it for the python processes is that adjustments are inherited by children, so unless we did something like this anyway, all the leaf processes would end up with the reduced score too.

samuelhwilliams: oh interesting, good to know 👍

samuelhwilliams (approved):
I think we can give this a go, although it would be good to think about how we monitor the success of this to decide if we need to keep it or if it's not really made enough of a difference to keep.

risicle:
Passes tests in preview.

PR #778, opened 2023-11-15
Bump pypdf back down to 3.13 to fix bug

idavidmcdonald:
We are seeing in our logs:
'**** Error: File did not complete the page properly and may be damaged.' and a small number of letters when uploaded are coming back as invalid due to content outside the printable area with the red box for invalid content in the wrong place.

I've replicated the issue locally, and dropping this back down to 3.13 fixes the issue for one of the PDFs in question.


samuelhwilliams (approved):
fixes for me locally too 👍

PR #784, opened 2023-11-22
Create bilingual PDFs when needed

samuelhwilliams:
We started rendering bilingual templates properly in https://github.com/alphagov/notifications-template-preview/pull/780, but this only applies to letter previews - ie what users can see on the admin site.

When letters are actually submitted to send, the API sends a task `create-pdf-for-templated-letter` to template-preview via celery, which creates a PDF that is prepared and configured for DVLA's printers. This patch adds support to the printable letter PDFs for including welsh pages, if the template is bilingual.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,9 +1,11 @@
 import base64
 from io import BytesIO
+from typing import Literal
 
 import boto3
 import sentry_sdk
 from botocore.exceptions import ClientError as BotoClientError
+from celery import Task
 from flask import current_app
 from flask_weasyprint import HTML
 from notifications_utils import LETTER_MAX_PAGE_COUNT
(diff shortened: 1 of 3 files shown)


app/celery/tasks.py
```diff
@@ -110,36 +113,53 @@ def copy_s3_object(source_bucket, source_filename, target_bucket, target_filenam
...
         admin_base_url=current_app.config["LETTER_LOGO_URL"],
         logo_file_name=logo_filename,
+        language=language,
     )
     with current_app.test_request_context(""):
         html = HTML(string=str(template))
 
     try:
-        with sentry_sdk.start_span(op="function", description="weasyprint.HTML.write_pdf"):
+        with sentry_sdk.start_span(op="function", description="weasyprint.HTML.write_pdf[english]"):
```
CrystalPea, line 131: should the language here be passed in, instead of hard-coded to [english]?

CrystalPea (approved):
Looks good.

As we talked, it looks like quite a lot of duplication between here and /preview.pdf endpoint. I will add exploring DRYing it to the refactor card 👍🏼

app/celery/tasks.py
samuelhwilliams: oops yes 🙌

PR #785, opened 2023-11-30
Dockerfile: increase `choom-wrap-packages` adjustment to 150

risicle:
The initial choice of 50 (in #774) was arbitrary and conservative, still seeing similar app deaths so let's try increasing it.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,7 +19,7 @@ RUN echo "Install binary app dependencies" \
         fonts-freefont-ttf=20120503-10 \
         fonts-wqy-zenhei \
         make \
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
No idea on whether 150 is a good second guess, but let's give it a try!

We do also have the option of giving the app more memory. Is that of interest too?

risicle:
Maybe?

Could alternatively decrease the concurrency of a worker to reduce the memory it is likely to take and reduce the blast radius of lost jobs when we lose an instance.

idavidmcdonald:
Good idea too about decreasing the concurrency of a worker. Lets get this deployed (as you've done) and monitor to see what happens. We can try alternative approaches if we find this isn't working.

PR #787, opened 2023-12-01
Dockerfile: increase `choom-wrap-packages` adjustment to 300

risicle:
When this starts to have an effect we should begin to see 500s resulting from `subprocess` calls being killed (perhaps with exit code 137).


risicle:
Oh yes, I guess when I say 500s I mean unhandled exceptions in general.

Come to think.. I haven't noticed any sentry errors from celery workers in general - do we actually have it set up properly to catch them?

PR #790, opened 2023-12-07
Hide second notify tag on letters for print

samuelhwilliams:
When we print letters, we insert a NOTIFY tag on the first page in the top-left, which helps DVLA printers recognise our workload.

We've been accidentally including this twice in bilingual letters, on the first page of Welsh and the first page of English content.

This patch hides the tag on the English pages.


CrystalPea (approved):
:shipit:

PR #793, opened 2024-01-02
notifications_utils: 74.0.0 -> 74.2.0

risicle:
https://github.com/alphagov/notifications-utils/compare/74.0.0...74.2.0


quis (approved):
Pro tip: you can run [`make bump-utils`](https://github.com/alphagov/notifications-template-preview/blob/a0952229ea28d67a5ae13b6e57228363b6aa8774/Makefile#L35-L37) to do this automatically and generate a nice commit message

PR #804, opened 2024-02-06
Update utils version to always hide barcodes in letters

CrystalPea:
So far we have been only hiding them when adding a NOTIFY tag.

This caused trouble with bilingual letters.

We tested the change on local and it hides the barcodes for printing.

See: https://github.com/alphagov/notifications-utils/pull/1096 for more details


whpearson (approved):
Changes look good

PR #815, opened 2024-03-12
PaaS Decommissioning

corlettb:
What
----

PaaS Decommissioning

WARNING
----

Do not merge yet.


klssmith (approved):
Looks good to me. I noticed we're no longer using `deploy-exclude.lst` so that could also be deleted, but we stopped using that some time ago - it wasn't related to moving off the PaaS.

corlettb:
Updated.

klssmith (approved):

PR #829, opened 2024-05-02
Dry up prepare pdf

leohemsted:


app/celery/tasks.py
```diff
@@ -226,6 +205,15 @@ def create_pdf_for_templated_letter(self: Task, encoded_letter_data):
     )
 
 
+def _prepare_pdf(letter_details, self):
+    def create_pdf_for_letter(letter_details, language, include_tag) -> BytesIO:
```
CrystalPea, line 209: Now I remembered - didn't we want to change this nested function and the same in `preview.py` into lambdas?

app/preview.py
```diff
@@ -225,7 +217,7 @@ def get_html(json, language="english"):
 
 
 @sentry_sdk.trace
```
CrystalPea, line 219: I wonder - do we still use this sentry tracing? Doesn't have to be removed in this PR, but might be a good time to think if we still need it?

app/celery/tasks.py
leohemsted: no, ruff will fail checks against lambdas assigned to variables, saying nested functions are preferred

we'd previously started with lambdas, then switched to get tests to pass, so i squashed that out.

looking elsewhere, there's some discussion saying "If you need to assign the lambda to a name, use a def instead"

https://stackoverflow.com/a/134638


i guess we could inline it, for example:

```
def _prepare_pdf(letter_details, self):
    purpose = PDFPurpose.PRINT

    return generate_templated_pdf(
        letter_details,
        (
            lambda letter_details, language, include_tag: _create_pdf_for_letter(
                self, letter_details, language=language, include_notify_tag=include_tag
            )
        ),
        purpose,
    )
```

but i don't particularly love that. open to hear other thoughts tho!

app/preview.py
leohemsted: hmmm i think lets keep this in here - i think i'd like a wider conversation about sentry before we decide what to do with these 🤔

app/celery/tasks.py
CrystalPea: Ok, no worries then ☺️

app/preview.py
CrystalPea: Okee 👍🏼

CrystalPea (approved):

PR #836, opened 2024-05-29
Fix bug in add_redact_annot() by pulling newest PyMuPDF

CrystalPea:
A user wrote to us with an issue - they tried to upload a PDF letter, but the letter content on the first page got mangled - symbols replaced the actual letter content on the first page.

Debugging the issue, I realised the text gets mangled during the address block redaction process. I looked that we use PyMuPDF to redact the address block, and that [they have fixed a couple issues](https://github.com/pymupdf/PyMuPDF/blob/8ec240783f66a227446d38d1afe732663d44b555/changes.txt) with `add_redact_annot()` in newer releases.

So I pulled in the newest release and upon testing it fixed the issue.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -16,7 +16,7 @@ wand==0.5.9
 pypdf==3.13.0
 reportlab==3.6.13
 pdf2image==1.12.1
-PyMuPDF==1.22.5
+PyMuPDF==1.24.4
 WeasyPrint==59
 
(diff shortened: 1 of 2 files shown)


leohemsted (approved):
https://github.com/pymupdf/PyMuPDF/issues/3257 looks like it might have been this, fixed in 1.24.0

PR #848, opened 2024-09-04
Change port mapping to use the loopback address

klssmith:
Docker defaults to using `0.0.0.0`, the default route listening on all network interfaces, as opposed to the loopback address (`127.0.0.1`). This has security implications and we have been asked to change it.

### DIFF
--- a/scripts/run_with_docker.sh
+++ b/scripts/run_with_docker.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 DOCKER_IMAGE_NAME=notifications-template-preview
+PORT=6013
 
 source environment.sh
 
 if [[ "${@}" == "web" || "${@}" == "web-local" ]]; then
-  EXPOSED_PORTS="-e PORT=6013 -p 6013:6013"
+  EXPOSED_PORTS="-e PORT=${PORT} -p 127.0.0.1:${PORT}:${PORT}"
 else
   EXPOSED_PORTS=""
 fi


scripts/run_with_docker.sh
```diff
@@ -4,7 +4,7 @@ DOCKER_IMAGE_NAME=notifications-template-preview
 source environment.sh
 
 if [[ "${@}" == "web" || "${@}" == "web-local" ]]; then
-  EXPOSED_PORTS="-e PORT=6013 -p 6013:6013"
+  EXPOSED_PORTS="-e PORT=6013 -p 127.0.0.1:6013:6013"
```
tombye, line 7: Why doesn't the value sent to the `-p` flag use the `POST` env var? Maybe it can't but otherwise it would be a bit neater.

tombye (commented):
The approach makes sense, I just have one comment so I'm clearer about how the line changed works.

scripts/run_with_docker.sh
klssmith: Looks like we just didn't set `PORT` as an environment variable for this app, we hardcoded it instead. We're quite inconsistent with how we do things, for example for some apps we use an environment variable for the port, others (like here) we don't, and in antivirus this line lives in the Makefile instead of the script to run with Docker.

I didn't attempt to make them all consistent, but I could have a look at doing that.

klssmith: Now updated to set the `PORT` on line 3 so it's easier to see at a glance which port we're using

tombye (approved):
Thanks for addressing the comment. Now looks good to me 👍🏻

PR #860, opened 2024-11-18
remove empty newlines from text parsing

leohemsted:
[Simplify text parsing comparison to make extract functions simpler](https://github.com/alphagov/notifications-template-preview/pull/860/commits/47a8b163c724c8747615ce3b9fb1625adeab3a99) 
47a8b16

After parsing the address, we just pass it in to the PostalAddress class, which does a bunch of address normalisation. We shouldn't waste our time doing this twice. So lets just jump ahead and do that step, then compare the rich address objects that are generated to assert that two precompiled letters would have exactly the same output


app/precompiled.py
```diff
@@ -713,8 +713,13 @@ def _extract_text_from_page(page, rect):
...
     def _get_address_from_get_textwords():
         text = page.get_text(clip=rect).strip()
-        # split and rejoin each line to remove any duplicate or trailing whitespace
-        return "\n".join(" ".join(line.split()) for line in text.split("\n"))
+        return "\n".join(
+            # split and rejoin each line to remove any duplicate or trailing whitespace
+            " ".join(line.split())
+            for line in text.split("\n")
+            # remove any lines that only contain empty strings
+            if line.strip()
```
quis, line 721: Not sure if it matters for this use case but it would be more robust to use `get_lines_with_normalised_whitespace`<sup>1</sup> from utils:

```suggestion
            line for line in get_lines_with_normalised_whitespace(text)
            # remove any lines that only contain empty strings
            if line.strip()
```

***

1. https://github.com/alphagov/notifications-utils/blob/4741402d194bb0dfc87d15b2dae16dd193cb5591/notifications_utils/formatters.py#L251C5-L252

leohemsted: good thought - you made me realise i could rework this - because i dont really care what this function returns, as long as when we put it into a `PrecompiledPostalAddress` that address matches. So splitting words onto diff lines because they're in diff fonts is bad. trailing whitespace, who cares if we're deleting this

https://github.com/alphagov/notifications-template-preview/commit/47a8b163c724c8747615ce3b9fb1625adeab3a99

app/precompiled.py
```diff
@@ -723,7 +721,9 @@ def _get_address_from_get_textwords():
         extracted_text.append(" ".join(w[4] for w in gwords))
     extracted_text = "\n".join(extracted_text)
 
-    if rect != NOTIFY_TAG_BOUNDING_BOX and _get_address_from_get_textwords() != extracted_text:
+    if rect != NOTIFY_TAG_BOUNDING_BOX and PrecompiledPostalAddress(
+        _get_address_from_get_textwords()
+    ) != PrecompiledPostalAddress(extracted_text):
```
quis, line 726: Will call through to `PostalAddress.__eq__` which looks sensible:
https://github.com/alphagov/notifications-utils/blob/4741402d194bb0dfc87d15b2dae16dd193cb5591/notifications_utils/recipient_validation/postal_address.py#L54-L63

quis (approved):
I get what you’re doing now, this feels like a good way of doing it

PR #876, opened 2025-03-31
log when filesize grows during sanitising

leohemsted:
we sometimes see issues where our system inflates the size of precompiled letters and attachments. the api checks that the filesize is below 2mb, but then we sometimes take a good pdf and make it too big at this point, causing technical failures and issues at DVLA which are hard to unpick.

while we work on fixing the underlying issues, here's some logging so we can at least be aware of how much it's happening.

i've set these to error logs if it's over 2mb or warning if it's still in size but did inflate. we might want to tweak these levels later if the logs are either not noisy enough or too noisy

### DIFF
--- a/app/precompiled.py
+++ b/app/precompiled.py
@@ -89,6 +89,9 @@ LOGO_BOTTOM_FROM_TOP_OF_PAGE = 30.00
 
 A4_HEIGHT_IN_PTS = A4_HEIGHT * mm
 
+MAX_FILESIZE = 2 * 1024 * 1024  # 2MB
+ALLOWED_FILESIZE_INFLATION_PERCENTAGE = 50  # warn if filesize after sanitising has grown by more than 50%
+
 precompiled_blueprint = Blueprint("precompiled_blueprint", __name__)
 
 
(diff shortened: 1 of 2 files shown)


app/precompiled.py
```diff
@@ -182,6 +185,27 @@ def sanitise_precompiled_letter():
     return jsonify(sanitise_json), status_code
 
 
+def _warn_if_filesize_has_grown(*, orig_filesize: int, new_filesize: int) -> None:
+    orig_kb = orig_filesize / 1024
+    new_kb = new_filesize / 1024
+
+    if new_filesize > MAX_FILESIZE:
+        current_app.logger.error(
+            "template-preview post-sanitise filesize too big: orig_size=%iKb; new_size=%iKb, over max_filesize=%iMb",
```
spatel033, line 194: Since we're adding this log, it would be helpful to include the file_name (notification_id.pdf) so it's easier to trace files during investigation or testing.

leohemsted: good suggestion

spatel033 (approved):

PR #878, opened 2025-04-24
Add celery critical logs and warnings as json

corlettb:


risicle (approved):
Let's try it I suppose

PR #886, opened 2025-06-06
Fix empty test assertions

quis:
`called_once_with` and `called_once` are not methods of `unittest.Mock`.

Instead they are mocked functions which will:
- accept any arguments
- always return something `True`-ish

This means the assertions in these tests are doing nothing, and changing their arguments or the arguments of the methods they are mocking does not cause the tests to fail.

This commit switches to using methods which do exist, and updating the arguments to make the tests pass.


risicle:
Famous bug, this is why I prefer comparing `.mock_calls` with a list of my expected `mock.call(...)` objects.

PR #891, opened 2025-06-25
Don’t check page exists twice for PNG preview

quis:
The `png_from_pdf` function checks that the requested page exists: https://github.com/alphagov/notifications-template-preview/blob/f3f3051071b8ef3d7ae11e39e044bf5c0e2aabb2/app/preview.py#L37-L40

We make the same check further up the stack here: https://github.com/alphagov/notifications-template-preview/blob/f3f3051071b8ef3d7ae11e39e044bf5c0e2aabb2/app/preview.py#L147-L155

`get_page_count_for_pdf` is slow as shit:
image

So let’s not do that.


CrystalPea (approved):
LGTM 👍🏼 

To review, I:
- looked at the code
- run the code locally and previewed letter templates and notification preview, also with letter attachment.

PR #892, opened 2025-06-30
Cache individual pages better

quis:
This pull request changes how we cache images of letter pages.

# Before

The cache key looked like:

> hash([contents of pdf] + [page number n])

Even the smallest change to the contents of the PDF would invalidate the cache for all images. For example filling in the address on page 1 meant page 10 would need to be re-rendered.

# After 

> hash([contents of page n])

The cache is only invalidated for pages which have changed. If editing something on page 1 doesn’t cause a reflow, page 10 is still cached.

***

This is especially advantageous where we have letters with attachments. Attachments are static, but they often contain graphics which are slow to render to PNGs.

Extracting the page content is work we have to do anyway in the case of a cache miss. In the case of a cache hit I estimate extracting and hashing a page will add about 150ms. This performance decrease is offset by the increased likelihood of a page being cached. It should also mean the simultaneous load on template preview is reduced when someone is clicking through and filling out the personalisation in a letter.

### DIFF
--- a/app/preview.py
+++ b/app/preview.py
@@ -107,7 +107,7 @@ def view_letter_template_png():
     # get pdf that can be read multiple times - unlike StreamingBody from boto that can only be read once
     requested_page = int(request.args.get("page", 1))
     pdf_persist = BytesIO(pdf) if isinstance(pdf, bytes) else BytesIO(pdf.read())
-    png_preview = get_png(
+    png_preview = png_from_pdf(
         pdf_persist,
         requested_page,
     )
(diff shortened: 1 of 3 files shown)


quis:
Update, found a consistent representation of the page data we can hash against, see 8e814fd

risicle:
I take it it doesn't work if you try to use `page_data_for_cache_key` as the actual stream to render?

quis:
@risicle nope, `page_data_for_cache_key` only contained the data for the page container. So it caused similar pages to get cached against the same key 😨 

***

Some quick performance testing trying to render page 1 of [example_dwp_pdf.pdf](https://github.com/alphagov/notifications-template-preview/blob/4b21b585801d802576d85d347b5391fd2ad3b309/tests/test_pdfs/example_dwp_pdf.pdf):

### Old (rasterize all pages)

> 1 loop, best of 5: 444 msec per loop

### New (extract requested page and rasterize)

> 1 loop, best of 5: 311 msec per loop

So not a massive advantage. I think unless we can cache against the actual PDF data (not a rasterized version) this isn’t worth it

risicle (approved):

PR #893, opened 2025-07-03
Wrap S3 with an in-memory cache

quis:
Once we put an attachment or cached PDF in S3 we never modify it. A new attachment or changes to a letter always results in a new cache object with a new key.

Each call to S3 takes ~50ms in production.

For viewing a 10 page letter in the admin app we need to:
- get the PDF to count the pages (1)
- get the PDF to extract each page to turn into a PNG (10)

That’s a total of 11 calls to S3 for the exact same PDF data.

Another case would be filling out the personalisation for a letter with an attachment. For each piece of personalisation we need to fetch the attachment PDF to merge it with the newly-generated templated letter before we cache the whole lot. So for a letter with 5 pieces of personalisation, that’s 5 repeated calls to S3 in fairly quick succession.

This commit puts an in-memory cache around calls to S3. So calls for the same object have a better chance of not going out to the network. 

***

It’s hard to know how effective this is without instrumentation.

Given 3 ECS tasks and 4 processes per task on production we think that means 12 separate caches.

With some simple modelling my guess is if you make 10 requests for the same thing in fairly quick succession you’ll have, on average, a 20% hit rate on the in-memory cache, with no performance penalty for a cache miss.

***

I’ve put the size of the cache at 500 items. Given a maximum PDF size of 2Mb (the most we allow for an attachment) this gives a worst-case cache size of 1Gb. I don’t know how much memory each worker has but they are only using about 10% of it at the moment:

image

***

Some typical request times from the browser console for loading a single PNG running locally. This exaggerates the improvement because my laptop isn’t very close to the S3 server. 

With memory cache (this branch) | Without memory cache (`main`)
---|---
61ms | 965ms
42ms | 743ms
71ms | 955ms
23ms | 935ms
55ms | 772ms

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -12,9 +12,10 @@ from notifications_utils.celery import NotifyCelery
 from notifications_utils.clients.signing.signing_client import Signing
 from notifications_utils.clients.statsd.statsd_client import StatsdClient
 from notifications_utils.logging import flask as utils_logging
-from notifications_utils.s3 import S3ObjectNotFound, s3download, s3upload
+from notifications_utils.s3 import S3ObjectNotFound, s3upload
 
 from app import weasyprint_hack
+from app.utils import caching_s3download
 
 notify_celery = NotifyCelery()
 metrics = GDSMetrics()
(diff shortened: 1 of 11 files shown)


CrystalPea (approved):
This looks good to me.

To review it, I:
- looked at the code
- have run the code locally and previewed a bilingual letter with attachment

PR #904, opened 2025-09-17
Upgrade docker version to trixie

spatel033:
- Changed the base image in the Dockerfile from debian bullseye to debian trixie
- Updated the pinned version of the required package in the Dockerfile to the latest version available in trixie
- Confirmed that the Docker image builds successfully
- Spun up the app locally/dev and verified that it converts to CMYK correctly
Deployment plan and rollback steps on this [card](https://trello.com/c/iUzmxrq9/1382-update-template-preview-docker-image-to-use-debian-trixie)

[PR](https://github.com/alphagov/notifications-template-preview/pull/907) for rollback or re-sanitise letter attachments task.

***

Closes https://github.com/alphagov/notifications-template-preview/pull/900


app/celery/tasks.py
```diff
@@ -269,3 +270,49 @@ def recreate_pdf_for_precompiled_letter(notification_id, file_location, allow_in
             notification_id,
         )
         return
+
+
+@notify_celery.task(name="recreate-pdf-for-template-letter-attachments")
+def recreate_pdf_for_template_letter_attachments(service_id, attachment_id, original_filename):
```
CrystalPea, line 276: To confirm if I understand correctly: we add this task, so if this deploy fails and we have to roll back and re-process some letters, we can re-sanitise the letter attachments?

In that case, shouldn't this new task be in a separate PR, deployed before the Trixie update? So that we can still access it if we roll back Trixie update PR?

PR #907, opened 2025-10-02
Add a task to re-sanitize the letter template attachment

spatel033:
We have added a new task called  to re-sanitize letter attachment PDF files. This task is triggered by notify-api.

Every time an attachment is uploaded, we keep an original copy in the letters-precompiled-originals-backup S3 bucket. The filename is the attachment ID. These files have already passed the virus scan, but they haven’t been sanitized yet. To send them to DVLA, they must be sanitized using template preview and then copied into the production-letter-attachments S3 bucket.

When we re-sanitize attachments, the process is:

- Get the file from the backup bucket.
- Sanitize the PDF using template preview.
- Upload it to the letter-attachments bucket with the correct metadata. If a file with the same name already exists, it will be replaced (overwritten).

We are logging the attachment ID if there’s an error during the process. This helps us trace the specific file, but it also means any failure will need manual investigation to resolve.


CrystalPea (commented):
Nice one! Could you add context in the commit message that we need this task as a safety measure - so if software updates lead to problems with sanitisation, we can roll back and re-sanitise?

It will make it easier in the future for someone with no context on this 🙌🏼 .

Other than that this is good to go 👍🏼

spatel033:
> Nice one! Could you add context in the commit message that we need this task as a safety measure - so if software updates lead to problems with sanitisation, we can roll back and re-sanitise?
> 
> It will make it easier in the future for someone with no context on this 🙌🏼 .
> 
> Other than that this is good to go 👍🏼

Thanks for review @CrystalPea, updated the commit message.

CrystalPea (approved):

PR #968, opened 2026-04-01
Re-freeze non-PDF and non-Celery requirements

quis:
This gives us up-to-date versions of dependencies which we think don’t need manual testing in order to change.

I’ve tried to organise the `requirements.in` file into logical sections to make it clearer what is/isn’t pinned.

There are many more things pinned to exact versions than in our other repos because of the problems we’ve had with PDF generation.

***

Also includes https://github.com/alphagov/notifications-template-preview/pull/958 because I think that’s a PDF-related dependency which is safe to update, and resolves a dependabot alert.


BlessedDev (approved):
Looks good to me, especially the `organisation of the requirements.in` file

PR #972, opened 2026-05-01
Ensure Arial is not distributed with the Dockerfile

klssmith:
We need Arial installed, but shouldn't distribute the font file directly in the Docker image. This removes the Arial.ttf file and installs it via `ttf-mscorefonts-installer` instead.

This package lives in Debian’s `contrib` component, so we need to enable `contrib` in `debian.sources` in order to install it.

[Trello card](https://trello.com/c/puNEyjD3/1784-ensure-we-are-not-publicly-distributing-arial)


docker/Dockerfile
```diff
@@ -11,7 +11,8 @@ ENV VIRTUAL_ENV="/opt/venv"
 COPY --chmod=0755 docker/choom-wrap-pkgs.sh /usr/local/bin/choom-wrap-pkgs
 
 RUN echo "Install binary app dependencies" \
-    && apt-get update && \
+    && sed -i 's/^Components: main/Components: main contrib/' /etc/apt/sources.list.d/debian.sources && \
```
ipdigital24, line 15: it may be useful to leave a comment on the need of this cmd
`sed -i 's/^Components: main/Components: main contrib/' /etc/apt/sources.list.d/debian.sources`

klssmith: Comment now added

ipdigital24 (approved):

PR #994, opened 2026-06-25
Upgrade Sentry from 1.45.1 to 2.63.0 

klssmith:
The migration guide can be seen here https://github.com/getsentry/sentry-python/blob/master/MIGRATION_GUIDE.md


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # This file was autogenerated by uv via the following command:
-#    uv pip compile requirements.in -o requirements.txt --exclude-newer 7d
```
ipdigital24, line 2: any reasons for removing `--exclude-newer 7d`

requirements_for_test.txt
```diff
@@ -1,5 +1,5 @@
 # This file was autogenerated by uv via the following command:
-#    uv pip compile requirements_for_test.in -o requirements_for_test.txt --exclude-newer 7d
```
ipdigital24, line 2: and here too

requirements.txt
klssmith: That flag gets added if you generate the requirements files by running `make refreeze-requirements`, but not if you run `make freeze-requirements`.  

So updating a top-level dependency (Sentry in this case), then running `make freeze-requirements` removes it automatically. But if someone is re-freezing requirements it will be added to the requirements files.

quis: https://github.com/alphagov/notifications-admin/pull/5976

PR #995, opened 2026-06-30
Update licence

ipdigital24:


LICENCE
```diff
@@ -1,4 +1,4 @@
-MIT License
+The MIT License (MIT)
 
 Copyright (c) 2017 Till Riedel
```
spatel033, line 3: Is this needed?

ipdigital24: Yes. I asked this question previously and understand that we have initially referred the code written by Till Riedel

spatel033 (approved):

PR #1001, opened 2026-07-13
Add notify tag area encroachment check

BlessedDev:
This PR addresses the recent incidents where the Notify tag has been obscured by encroaching whitespace which is not visible and resulted in the affected letters not being printed by [DVLA](https://trello.com/c/pniAGVmJ/685-check-for-invisible-characters-in-notify-tag-area-for-precompiled-letters).

I have left extensive comments in the PR to explain the decisions made in the PR.

I also tested the code with pymupdf version 1.28.0, the current version is 1.24.11 but there is upcoming work to bump the version. It works with either version.


app/precompiled.py
```diff
@@ -261,6 +261,21 @@ def sanitise_file_contents(encoded_string, *, allow_international_letters, filen
...
+            file_name = filename
+            encroaching_text = encroaching_text
+            current_app.logger.exception(
+                "precompiled pdf:(%s) has character: (%s), encroaching on the Notify tag area.",
+                file_name,
+                encroaching_text,
+                extra={"file_name": filename, "encroaching_text": encroaching_text},
+            )
+
+            raise ValidationFailed(message, page_count=page_count)
```
spatel033: We can deploy this in two phases. Instead of raising an error initially, we can log the error, monitor the logs, and cancel the letters if they contain invalid characters.

If valid letters are not failing, we can then enable the validation failure error.

app/precompiled.py
```diff
@@ -826,6 +841,77 @@ def is_notify_tag_present(pdf):
...
+    return None
+
+
+def _no_intersect_with_notify_tag_bbox(bbox):
+    """
+    Returns True if a bbox does not intersect with NOTIFY_TAG_BOUNDING_BOX in any way.
+    """
+    t_x0 = NOTIFY_TAG_BOUNDING_BOX.x0
+    t_y0 = NOTIFY_TAG_BOUNDING_BOX.y0
+    t_x1 = NOTIFY_TAG_BOUNDING_BOX.x1
```
spatel033: Currently, the validation only checks within the `NOTIFY_TAG_BOUNDING_BOX_WIDTH` area. If the any characters are outside this area, the validation will pass.
Image

`t_x1` should be `A4_WIDTH * mm` to check for characters across the full width of the page.

Image

app/precompiled.py
BlessedDev: That is a good catch, I will address the invalid_pages issue.
Won't it cause issues if a user uploads a precompiled letter with no validation error only to find out that the letter was cancelled due to a validation error?

app/precompiled.py
BlessedDev: The check is only deliberately concentrated on the Notify tag area and will also pick up any bounding boxes that intersects with it. My understanding was that the DVLA issues where down to the characters encroaching on the  Notify tag area. Does the issue also occur if a character is any where across the width of the page, even if doesn't encroach on the NOTIFY_TAG_BOUNDING_BOX area?

app/precompiled.py
rparke: It's still an improvement on what we have now where they go into technical failure AFTER a user thinks they've been sent to print.

app/precompiled.py
```diff
@@ -826,6 +827,73 @@ def is_notify_tag_present(pdf):
...
+
+                # Account for the fact that the text "NOTIFY" is in the notify_tag bounding box
+                if "NOTIFY" in text:
+                    text = text.replace("NOTIFY", "")
+
+                # Any remaining text or trailing ghost spaces will trigger an encroachment
+                if text:
+                    return {"result": True, "text": text}
+
+    return {"result": False, "text": None}
```
rparke: Are we using this dict as a dict anywhere else?

If not we can just return false here and return the dict if we have a result (could wrap it in an object if you wanted to do more rigorous checks)

app/precompiled.py
BlessedDev: validation error

I have addressed the issue raised. This will be the validation error presented to the user, I appreciate they wouldn't be able to do much because it is unlikely they will see the invisible text. However the details of the error is being logged and the user will be clear that the letter wasn't sent to print.

app/precompiled.py
BlessedDev: This has been addressed as well. The scan now encompasses the entire width of the page to the right of the notify tag bounding box. It should address the type of incident where invisible text outside the notify tag area but to the right of it resulted in issues at DVLA.

app/precompiled.py
BlessedDev: No, that was a hold over and I had forgotten to push the updated commit. It's been resolved now.

app/precompiled.py
```diff
@@ -261,6 +261,21 @@ def sanitise_file_contents(encoded_string, *, allow_international_letters, filen
...
             )
 
+        # Check if there are encroaching invisible/hidden characters on the Notify tag area
+        encroaching_text = check_notify_tag_area_for_encroachment(file_data)
+        if encroaching_text:
+            message = "content-outside-printable-area"
+            file_name = filename
+            encroaching_text = encroaching_text
+            current_app.logger.exception(
+                "precompiled pdf:(%s) has character: (%s), encroaching on the Notify tag area.",
```
klssmith: It would be best not to log the exact text in case it contains PII, in the same way that we deliberately don't log any non-compatible characters in SMS, we only log that they're there. We can still find see the original file if we need to see what was there.

We also might not want this to be at a lower level than an exception, since the validation error is caught and handled and there's nothing for us to investigate. We'll still be able to check for this happening

BlessedDev: I will skip logging the exact text, however `check_notify_tag_area_for_encroachment` will still return the encroaching text in order to support the unit tests that ensure that a broad range of invisible/hidden characters are being detected.

BlessedDev: For now I have split the difference by handling the validation error as other `content-outside-printable-area` validation errors when uploading precompiled letters and logging the event as an exception.
We can observe how this works when deployed and make adjustments.

